PR #4, opened 2020-06-08
Make example code for service enablement work

corvus-ch:
The for loop did not yield the expected results.


docs/modules/ROOT/pages/how-tos/install/gcp.adoc
```diff
@@ -110,11 +110,11 @@ roles/storage.admin
 roles/dns.admin
 EOF
 
-for role in $roles; do
+while IFS= read -r role; do
```
srueg, line 113: What's wrong with the `for` loop? Seems to work just fine?

PR #6, opened 2020-06-08
Mention some handy tools

corvus-ch:
Mention Purser and the "OpenShift Limits Calculator" in the GCP install how to.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/install/gcp.adoc
+++ b/docs/modules/ROOT/pages/how-tos/install/gcp.adoc
@@ -90,7 +90,7 @@ NOTE: Plan ahead as this can take up to two days. It usually takes only seconds 
 [TIP]
 ====
 Use the https://access.redhat.com/labs/ocplimitscalculator/[OpenShift Limits Calculator] to calculate your needs.
-This is specialy handy when it is planned to setup several clusters within one project.
+This is especially handy when it is planned to setup several clusters within one project.
 It does not take regions into consideration and some regions have different defaults.
 Take the column "Service Limit increase necessary?" with a grain of salt and double check.
 ====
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/install/gcp.adoc
```diff
@@ -86,6 +86,14 @@ This might not be the case when using an account already containing resources.
 Raising quotas can be done at https://console.cloud.google.com/iam-admin/quotas.
 +  
 NOTE: Plan ahead as this can take up to two days. It usually takes only seconds though.
++
+[TIP]
+====
+Use the https://access.redhat.com/labs/ocplimitscalculator/[OpenShift Limits Calculator] to calculate your needs.
+This is specialy handy when it is planned to setup several clusters within one project.
```
srueg, line 93:
```suggestion
This is especially handy when it is planned to setup several clusters within one project.
```

docs/modules/ROOT/pages/how-tos/install/gcp.adoc
```diff
@@ -148,6 +156,8 @@ Prerequisites:
 * GCP project setup according to above documentation
 * Service Account Key (key.json)
 
+TIP: https://github.com/projectsyn/purser[Purser] can be used to check if the preconditons are met.
```
tobru, line 159: I would wait a bit before including this one. For me it's still unclear if we have to invent this cloud-checker tool ourselves...

corvus-ch: Conclusion from a discussion between @tobru, @srueg and me:

Even if we find a utility built for the same/a similar use case, it is likely that we need something to make usage easy. Purser can/will be that wrapper. We agreed on keeping it mentioned here.

PR #23, opened 2020-07-20
Commodore Component Developer Guidelines

srueg:
Closes https://ticket.vshn.net/browse/APPU-2758

### DIFF
--- a/docs/modules/ROOT/pages/references/projectsyn/developer.adoc
+++ b/docs/modules/ROOT/pages/references/projectsyn/developer.adoc
@@ -1,1 +1,136 @@
 = Commodore Component Developer Guidelines
+
+This page defines the various conventions and best practices for writing a https://syn.tools/commodore/index.html[Commodore component] for OpenShift 4.
+
+[NOTE]
+====
+See the https://syn.tools/commodore/writing-a-component.html[Writing a Commodore Component] guide for a more general introduction and technical details of Commodore components.
+====
+
+== Conventions Used in These Guidelines
+
+The requirement level keywords _must_, _must not_, _required_, _shall_, _shall not_, _should_, _shouldn't_, _recommended_, _may_, and _optional_ used in this document are to be interpreted as described in https://www.ietf.org/rfc/rfc2119.txt[RFC 2119].
+
+
+== Component Naming
+
+The name of a component which configures something OpenShift 4 specific _must_ be prefixed with `openshift4-`.
+
+[NOTE]
+====
+When using the `commodore component new` command to bootstrap a new component, the `component-` prefix will be added automatically and _must not_ be specified:
+
+[source,console]
+----
+commodore component new openshift4-registry --name "OpenShift 4 Registry" --owner appuio
+----
+====
+
+
+== https://docs.openshift.com/container-platform/latest/operators/understanding_olm/olm-understanding-olm.html[Operator Lifecycle Manager (OLM)]
+
+We _recommend_ to avoid using the OLM to install operators on a cluster.
+This is mainly due to the added complexity the OLM introduces and us not using any of it's features. This might change in the future.
+
+Exceptions to this rule include cases where the Red Hat documentation explicitly requires the OLM.
+
+Any operators we **do** install with the OLM _should_ be restricted to single namespaces and _must not_ be installed using the `AllNamespaces` InstallMode.
+
+We don't provide any support for operators installed by customers via the OLM.
+
+
+== CustomResource (CR) Specs
+
+Components which configure one or multiple CRs _should_ do so by exposing its `.spec` field as a parameter.
+This enables the configuration of all possible fields via the inventory and provides the maximum flexibility.
+Default values _may_ either be set in the `defaults.yml` file of the component or in Jsonnet code.
+It _should_ always be possible to override default values in the inventory.
+
+[CAUTION]
+====
+Lists in the inventory get extended by http://reclass.pantsfullofunix.net/operations.html#data-merging[reclass] when being merged together.
+This means that the items of each list are added together in a new list.
+It can lead to undesired behavior if not accounted for.
+====
+
+If creating multiple CRs, a dict _should_ be used where its keys map to the names used for the CRs.
+Iterating over the keys of this dict in Jsonnet enables the enrichment of possible default values if not set.
+
+Due to the way http://reclass.pantsfullofunix.net/index.html[reclass] handels `null` overrides, the component _must_ handle `null` values in dicts and lists gracefully.
+
+.main.jsonnet
+[source,jsonnet]
+----
+local ingressControllers =
+  if params.ingressControllers != null
+  then std.objectFields(params.ingressControllers)
+  else [];
+
+{
+  [name]:
+    [kube._Object('operator.openshift.io/v1', 'IngressController', name) {
+      metadata+: {
+        namespace: params.namespace,
+      },
+      spec: {
+        defaultCertificate: {
+          name: acmeCertName,
+        },
+      } + params.ingressControllers[name],
+    }]
+  for name in ingressControllers
+} + {
+  [if std.length(ingressControllers) == 0 then '.gitkeep']: {},
+}
+----
+
+
+== Inventory Parameters
+
+A component _may_ define various parameters in the inventory (`defaults.yml`).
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,125 @@
...
+Any operators we **do** install with the OLM should be restricted to single namespaces and must never be installed using the `AllNamespaces` InstallMode.
+
+We don't provide any support for operators installed by customers via the OLM.
+
+
+== CustomResource (CR) Specs
+
+Components which configure one or multiple CRs should do so by exposing its `.spec` field as a parameter.
+This enables the configuration of all possible fields via the inventory and provides the maximum flexibility.
+Default values can either be set in the `defaults.yml` file of the component or in JSONNET code.
```
corvus-ch, line 42:
```suggestion
Default values can either be set in the `defaults.yml` file of the component or in Jsonnet code.
```

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,125 @@
...
+It should always be possible to override default values in the inventory.
+
+If creating multiple CRs, a dict should be used where its keys map to the names used for the CRs.
+Iterating over the keys of this dict in JSONNET enables the enrichment of possible default values if not set.
```
corvus-ch, line 46:
```suggestion
Iterating over the keys of this dict in Jsonnet enables the enrichment of possible default values if not set.
```

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,125 @@
...
+          name: acmeCertName,
+        },
+      } + params.ingressControllers[name],
+    }]
+  for name in ingressControllers
+} + {
+  [if std.length(ingressControllers) == 0 then '.gitkeep']: {},
+}
+----
+
```
corvus-ch, line 75: Please also hint the issue with arrays. While we do not have a solution, I deem it important to have it mentioned here. Component authors can then lookout for those cases and see if they have to mitigate them or not.

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+== Inventory Parameters
+
+A component can define various parameters in the inventory (`defaults.yml`).
```
corvus-ch, line 79: This is not OpenShift specific and duplicates what would have do be documented by Syn. Assuming the component got bootstrapped with `commodore new component`, this is already a given.

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,125 @@
...
+    namespace: openshift-ingress
+    cloud:
+      provider: ${cloud:provider}
+    ingressControllers:
+      default:
+        replicas: 2
+
+----
+
+If the component requires additional information like the cloud provider, cluster version or infrastructure ID, these can also be accessed via parameters of the inventory.
```
corvus-ch, line 96: Make it clear that this is about facts—or the lack thereof. For me this currently reads confusing.

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
srueg: Yes, my thoughts exactly. But I thought better have it documented here first rather than nowhere. We can always move it again. WDYT?

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
srueg: Do you mean the issue regarding merging of arrays in the inventory?

corvus-ch: Yes.

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
corvus-ch: In that case, I would more focus on using the `new component` command than explaining how the naming is supposed to be. Or at least say something along the line: use that and this is solved for you.

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,136 @@
...
+
+Any operators we **do** install with the OLM _should_ be restricted to single namespaces and _must not_ be installed using the `AllNamespaces` InstallMode.
+
+We don't provide any support for operators installed by customers via the OLM.
+
+
+== CustomResource (CR) Specs
+
+Components which configure one or multiple CRs _should_ do so by exposing its `.spec` field as a parameter.
+This enables the configuration of all possible fields via the inventory and provides the maximum flexibility.
```
corvus-ch, line 45:
```suggestion
This enables the configuration of all possible fields via the inventory and provides maximum flexibility.
```

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,136 @@
...
+It _should_ always be possible to override default values in the inventory.
+
+[CAUTION]
+====
+Lists in the inventory get extended by http://reclass.pantsfullofunix.net/operations.html#data-merging[reclass] when being merged together.
+This means that the items of each list are added together in a new list.
+It can lead to undesired behavior if not accounted for.
+====
+
+If creating multiple CRs, a dict _should_ be used where its keys map to the names used for the CRs.
```
corvus-ch, line 56:
```suggestion
When creating multiple CRs, a dict _should_ be used where its keys map to the names used for the CRs.
```

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,136 @@
...
+Iterating over the keys of this dict in Jsonnet enables the enrichment of possible default values if not set.
```
corvus-ch, line 57:
```suggestion
Iterating over the keys of this dict in Jsonnet enables the enrichment of possible default values.
This allows to omit default values within the inventory.
```

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,136 @@
...
+} + {
+  [if std.length(ingressControllers) == 0 then '.gitkeep']: {},
+}
+----
+
+
+== Inventory Parameters
+
+A component _may_ define various parameters in the inventory (`defaults.yml`).
+These _should_ be namespaced in a dict with the component's name (hyphens replaced with underscores):
```
corvus-ch, line 91:
```suggestion
These _must_ be namespaced in a dict with the component's name (hyphens replaced with underscores):
```

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
```diff
@@ -0,0 +1,136 @@
...
+    cloud:
+      provider: ${cloud:provider}
+    ingressControllers:
+      default:
+        replicas: 2
+
+----
+
+If the component requires additional information like the cloud provider, cluster version or infrastructure ID, these _may_ also be accessed via parameters of the inventory (usually in the form of facts).
+In the future this information _should_ be automatically collected on the cluster and exposed in the inventory as facts.
```
corvus-ch, line 108:
```suggestion
In the future this information will be automatically collected on the cluster and exposed in the inventory as facts.
```

docs/modules/ROOT/pages/references/projectsyn/developer.adoc
srueg: I thought about this too but decided to stick with `should` because it might make sense to define other dicts as well. Especially for components which are tightly coupled like synsights or crossplane/crossplane-cloudscale.

corvus-ch: Both Synsights and Crossplane are early adopter components. Since then we have learned and I question what they are doing.

Mabybe we could go with _must_ but relax it by the following.

> A group of components _may_ agree upon a common prefix to share parameters between each other.

PR #35, opened 2020-08-31
Implement feedback for cloudscale.ch installation

srueg:
Closes https://ticket.vshn.net/browse/APPU-2849

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
@@ -6,7 +6,10 @@ Steps to install an OpenShift 4 cluster on https://cloudscale.ch[cloudscale.ch].
 
 These steps follow the https://docs.openshift.com/container-platform/latest/installing/installing_bare_metal/installing-bare-metal.html[Installing a cluster on bare metal] docs to set up a user provisioned installation (UPI).
 https://www.terraform.io[Terraform] is used to provision the cloud infrastructure.
+--
 
+[NOTE]
+--
 The commands are idempotent and can be retried if any of the steps fail.
 --
 
@@ -30,7 +33,14 @@ It's currently very specific to VSHN and needs further changes to be more generi
 
 == Cluster Installation
 
-Register the new OpenShift 4 cluster in Lieutenant: https://control.vshn.net/syn/lieutenantclusters
+Register the new OpenShift 4 cluster in Lieutenant.
+
+.Lietenant API endpoint
+****
+Use the following endpoint for Lieutenant:
+
+VSHN:: https://control.vshn.net/syn/lieutenantclusters
+****
 
 === Configure input
 
@@ -42,7 +52,7 @@ export GITLAB_CATALOG_PROJECT_ID=<project-id> # GitLab numerical project ID of t
 export REGION=rma # rma or lpg (without the zone number)
 export CLUSTER_ID=<lieutenant-cluster-id>
 export TENANT_ID=<lieutenant-tenant-id>
-export BASE_DOMAIN=ocp4-poc.appuio-beta.ch # See xref:explanation/dns_scheme.adoc[DNS Scheme]
+export BASE_DOMAIN=appuio-beta.ch # See xref:explanation/dns_scheme.adoc[DNS Scheme]
 export PULL_SECRET=<redhat-pull-secret> # As copied from https://cloud.redhat.com/openshift/install/pull-secret "Copy pull secret"
 ----
 
@@ -87,7 +97,7 @@ mc mb --ignore-existing \
 
 For the following steps, change into your local checkout of the https://github.com/projectsyn/commodore[Commodore] repository.
 
-[WARNING]
+[CAUTION]
 These are the only steps which aren't idempotent.
 If you have to recreate the install config or any of the generated manifests you need to rerun all of the subsequent steps.
 
@@ -251,7 +261,9 @@ EOF
 terraform apply
 ----
 
-. Create the shown DNS records and wait for them to propagate!
+. Create the shown DNS records
+
+. Wait for the DNS records to propagate!
 +
 [source,console]
 ----
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
@@ -6,6 +6,8 @@ Steps to install an OpenShift 4 cluster on https://cloudscale.ch[cloudscale.ch].
 
 These steps follow the https://docs.openshift.com/container-platform/latest/installing/installing_bare_metal/installing-bare-metal.html[Installing a cluster on bare metal] docs to set up a user provisioned installation (UPI).
 https://www.terraform.io[Terraform] is used to provision the cloud infrastructure.
+
+The commands are idempotent and can be retried if any of the steps fail.
 --
 
 [NOTE]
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -6,6 +6,8 @@ Steps to install an OpenShift 4 cluster on https://cloudscale.ch[cloudscale.ch].
 
 These steps follow the https://docs.openshift.com/container-platform/latest/installing/installing_bare_metal/installing-bare-metal.html[Installing a cluster on bare metal] docs to set up a user provisioned installation (UPI).
 https://www.terraform.io[Terraform] is used to provision the cloud infrastructure.
+
+The commands are idempotent and can be retried if any of the steps fail.
```
corvus-ch, line 10:
```suggestion
NOTE: The commands are idempotent and can be retried if any of the steps fail.
```

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -14,117 +16,157 @@ This how-to guide is still a work in progress and will change.
...
 * `yq` https://mikefarah.gitbook.io/yq[yq YAML processor]
 * `openshift-install` https://cloud.redhat.com/openshift/install/metal/user-provisioned[OpenShift Installer]
-* Working local https://github.com/projectsyn/commodore[Commodore] setup
+* https://github.com/projectsyn/commodore[Commodore] docker-compose setup
 
 
 == Cluster Installation
 
-. Register the new OpenShift 4 cluster in Lieutenant: https://control.vshn.net/syn/lieutenantclusters
+Register the new OpenShift 4 cluster in Lieutenant: https://control.vshn.net/syn/lieutenantclusters
```
corvus-ch, line 33: VSHN specifics must be placed in a sidebar.

```suggestion
Register the new OpenShift 4 cluster in Lieutenant.

.Lietenant API endpoint
****
Use the following endpoint for Lieutenant:

VSHN:: https://control.vshn.net/syn/lieutenantclusters
****
```

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -14,117 +16,157 @@ This how-to guide is still a work in progress and will change.
...
-export REGION=rma # rma or lpg
-export CLUSTER_ID=<cluster-id>
-export TENANT_ID=<tenant-id>
-export BASE_DOMAIN=ocp4-poc.appuio-beta.ch
-export PULL_SECRET=<redhat-pull-secret> # From https://cloud.redhat.com/openshift/install/pull-secret
+export GITLAB_CATALOG_PROJECT_ID=<project-id> # GitLab numerical project ID of the catalog repo
+export REGION=rma # rma or lpg (without the zone number)
+export CLUSTER_ID=<lieutenant-cluster-id>
+export TENANT_ID=<lieutenant-tenant-id>
+export BASE_DOMAIN=ocp4-poc.appuio-beta.ch # See xref:explanation/dns_scheme.adoc[DNS Scheme]
```
corvus-ch, line 45:
```suggestion
export BASE_DOMAIN=appuio-beta.ch # See xref:explanation/dns_scheme.adoc[DNS Scheme]
```

The subdomain `ocp4-poc` shall only be used for clusters placed within the corresponding GCP project.

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -14,117 +16,157 @@ This how-to guide is still a work in progress and will change.
...
-  "s3/${CLUSTER_ID}-bootstrap-ignition"
+mc mb --ignore-existing \
+  "${CLUSTER_ID}/${CLUSTER_ID}-bootstrap-ignition"
 ----
 
+=== OpenShift Installer Setup
+
+For the following steps, change into your local checkout of the https://github.com/projectsyn/commodore[Commodore] repository.
+
+[WARNING]
```
corvus-ch, line 90:
```suggestion
[CAUTION]
```

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -14,117 +16,157 @@ This how-to guide is still a work in progress and will change.
...
 platform:
   none: {}
 pullSecret: |
   ${PULL_SECRET}
-sshKey: "$(cat ~/.ssh/id_ed25519.pub)"
 EOF
 ----
 
-. Prepare install manifests and ignition config
+. Render install manifests (this will consume the `install-config.yaml`)
```
corvus-ch, line 112: For GCP, the install config gets written to a `<cluster id>-config` and then gets copied to `<cluster-id>`. By doing so, of copy of `install-config.yaml` (and some additional steps) will not get touched by the installer. They can be copied over and over again should there be a need.

It probably makes sense to apply the same pattern here.

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -198,7 +251,13 @@ EOF
 terraform apply
 ----
 
-. Create the shown DNS records
+. Create the shown DNS records and wait for them to propagate!
```
corvus-ch, line 254:
```suggestion
. Create the shown DNS records

 . Wait for the DNS records to propagate!
```

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -198,7 +251,13 @@ EOF
...
++
+[source,console]
+----
+sleep 600
+host "api.${CLUSTER_ID}.${BASE_DOMAIN}"
```
corvus-ch, line 259:
```suggestion
while -n 15 host "api.${CLUSTER_ID}.${BASE_DOMAIN}"
```

Not sure if that really helps. I have the habit to run such test commands using while.

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -14,117 +16,157 @@ This how-to guide is still a work in progress and will change.
...
-* `mc` https://docs.min.io/docs/minio-client-quickstart-guide.html[Minio client]
+
+* `docker`
+* `docker-compose`
+* `mc` https://docs.min.io/docs/minio-client-quickstart-guide.html[Minio client] (aliased to `mc` if necessary)
 * `jq`
 * `yq` https://mikefarah.gitbook.io/yq[yq YAML processor]
 * `openshift-install` https://cloud.redhat.com/openshift/install/metal/user-provisioned[OpenShift Installer]
-* Working local https://github.com/projectsyn/commodore[Commodore] setup
+* https://github.com/projectsyn/commodore[Commodore] docker-compose setup
```
tobru, line 28: Instead of Docker Compose, could something like https://syn.tools/tuto/index.html#_using_the_commodore_docker_image be used?

corvus-ch (changes requested):
Initially we intended to make this documentation as generic as possible. The idea was to make them work for others too (for example Puzzle).

Meanwhile, I have my doubts if this is good thing or not. Opinions: @tobru @srueg?

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
srueg: Yes and no: the setup of such a command would include quite some additional steps (duplicating the README) for API, Git URLs, credentials, etc.

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
srueg: I don't really see the benefit of this since it's a copy-paste command to create the installer-conifg.yaml anyway and it only contains env vars. So nothing really that needs to be retained?

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
srueg: Unfortunately the command doesn't work like this?

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
srueg: Maybe we should add a "local install" how-to to https://github.com/projectsyn/commodore/pull/147 ?

srueg:
> Initially we intended to make this documentation as generic as possible. The idea was to make them work for others too (for example Puzzle).
> 
> Meanwhile, I have my doubts if this is good thing or not. Opinions: @tobru @srueg?

Hence the NOTE at the top:
> It's currently very specific to VSHN and needs further changes to be more generic.

But I agree, eventually this should become more generic and non VSHN specific. Maybe not something to do right now though?

tobru:
> But I agree, eventually this should become more generic and non VSHN specific. Maybe not something to do right now though?

Exactly, at the current stage I don't know how to make this VSHN independent. But for sure this should be the end goal...

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
tobru: I don't like the fact that you need to "install" commodore when you can just do a "docker run" with some parameters. As this documentation already is very VSHN specific, why not add the VSHN specific copy/paste command for the time being? We might also document how to use the `./commodore_command.sh` "officially" and say that the `.env` file is the Commodored "configuration file", which we could provide as an example in our VSHN wiki?

srueg: Ok, I added the alias to use the docker image directly.

tobru (approved):

corvus-ch (approved):

PR #50, opened 2020-11-02
Update cloudscale installation documentation

ccremer:
Propose fixes for errors while doing the tutorial

Todo:
- [x] existing Tenant and Cluster are a prerequisite
- [x] make idempotent: duplicate applications line
- [x] split terraform alias and next block
- [x] extend mc expiration to 2 or 3 hours
- [x] Move _Set up LDAP service_ to (before) Configure input
- [x] Put direct link to OpenShift installer (version matching the documentation)
- [x] Drop "Create the remaining DNS records. Pay attention that service DNS records must end with a dot (.)". The output is already shown on the previous command.
- [x] `pwgen` in the requirements, or use `LC_ALL=C tr -cd "A-Za-z0-9" </dev/urandom | head -c 128`
- [x] `http` is also used, but missing in requirements. Replace with `curl`, which is more widespread: `curl -sH "${LIEUTENANT_AUTH}" "https://${LIEUTENANT_URL}/clusters/${CLUSTER_ID}" | jq -r ".installURL"`
- [x] LIEUTENANT_TOKEN is nowhere defined
- [x] Cluster catalog compilation succeeds the first time (at least in my case). Maybe remove the lines?

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
@@ -22,6 +22,11 @@ This how-to guide is still a work in progress and will change.
 It's currently very specific to VSHN and needs further changes to be more generic.
 --
 
+== Starting situation
+
+* You already have a Tenant and its git repository
+* You have a CCSP Red Hat login and are logged into https://cloud.redhat.com/openshift/install/metal/user-provisioned[Red Hat Openshift Cluster Manager]
+* You want to register a new cluster in Lieutenant and are about to install Openshift 4 on Cloudscale
 
 == Prerequisites
 
@@ -29,7 +34,7 @@ It's currently very specific to VSHN and needs further changes to be more generi
 * `mc` https://docs.min.io/docs/minio-client-quickstart-guide.html[Minio client] (aliased to `mc` if necessary)
 * `jq`
 * `yq` https://mikefarah.gitbook.io/yq[yq YAML processor]
-* `openshift-install` https://cloud.redhat.com/openshift/install/metal/user-provisioned[OpenShift Installer]
+* `openshift-install` https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.5/openshift-install-linux.tar.gz[OpenShift Installer]
 * `vault` https://www.vaultproject.io/docs/commands[Vault CLI]
 * `rhcos-4.5` image must be present in the cloudscale.ch account where you want to install
 
@@ -49,6 +54,20 @@ Use the following endpoint for Lieutenant:
 VSHN:: https://control.vshn.net/syn/lieutenantclusters
 ****
 
+=== Set up LDAP service
+
+. Create an LDAP service
++
+Use https://control.vshn.net/vshn/services/_create to create a service.
+The name must contain the customer and the cluster name.
+And then put the LDAP service ID in the following variable:
++
+[source,console]
+----
+export LDAP_ID="Your_LDAP_ID_here"
+export LDAP_PASSWORD="Your_LDAP_pw_here"
+----
+
 === Configure input
 
 [source,console]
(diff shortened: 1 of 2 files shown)


ccremer:
questions:
* To edit terraform docs, there's this `include::partial$setup_terraform.adoc[]`, but this file is not in this repo. Where is it?
* After I've partially restarted my failed attempt to install it from last week, I deleted the local folder, but not the git repository (there were commits etc. So, the step that said "this commodore install will fail" actually succeeded the first in my 2nd attempt. Do you have an idea if that is related to the first attempt, or has something changed so that it now always succeeds first time?

srueg:
* It's here: https://github.com/appuio/openshift4-docs/blob/master/docs/modules/ROOT/partials/setup_terraform.adoc
* If there still was Terraform state (stored on the GitLab repo) then yes, probably due to the first attempt...

ccremer:
>     * It's here: https://github.com/appuio/openshift4-docs/blob/master/docs/modules/ROOT/partials/setup_terraform.adoc

I was blind in VScode, and grep -r doesn't count the file name as usage itself :D

>     * If there still was Terraform state (stored on the GitLab repo) then yes, probably due to the first attempt...

I don't think it's about terraform, but the commodore catalog compilation that fails, so a step before that.

srueg:
> I don't think it's about terraform, but the commodore catalog compilation that fails, so a step before that.

Ah, you're right. But I think it will still fail for new clusters, because for example the cluster-id is not known before provisioning the cluster etc.

ccremer:
Ok thanks, I re-added the notice.

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -396,8 +401,9 @@ Install Steward on the cluster (see https://wiki.vshn.net/pages/viewpage.action?
 +
 [source,console]
 ----
-export LIEUTENANT_URL="api.syn.vshn.net"
-export LIEUTENANT_NS="lieutenant-prod"
+export LIEUTENANT_URL="api.syn.vshn.net" # same as COMMODORE_API_URL
+export LIEUTENANT_NS="lieutenant-prod" # or lieutenant-[dev,int] accordingly
+export LIEUTENANT_TOKEN=<lieutenant-api-token> # yes, it's the same as COMMODORE_API_TOKEN
 export LIEUTENANT_AUTH="Authorization:Bearer ${LIEUTENANT_TOKEN}"
```
corvus-ch, line 407: Why not just use `COMMODORE_API_TOKEN` here? The variable `LIEUTENANT_TOKEN` not used anywhere else and do not have to be set.

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -408,7 +414,7 @@ curl \
   -d '[{ "op": "remove", "path": "/status/bootstrapToken" }]' \
   "https://rancher.vshn.net/k8s/clusters/c-c6j2w/apis/syn.tools/v1alpha1/namespaces/${LIEUTENANT_NS}/clusters/${CLUSTER_ID}/status"
 
-kubectl apply -f $(http -pb "https://${LIEUTENANT_URL}/clusters/${CLUSTER_ID}" "${LIEUTENANT_AUTH}" | jq -r ".installURL")
+kubectl apply -f $(curl -sH "${LIEUTENANT_AUTH}" "https://${LIEUTENANT_URL}/clusters/${CLUSTER_ID}" | jq -r ".installURL")
```
corvus-ch, line 417: Can we set the `--kubeconfig` here? It happened to me several times, that the wrong Kubeconfig was used.

corvus-ch (changes requested):

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
srueg: It should be set by the `KUBECONFIG` env var of the previous step. But yes, setting it explicitly might be a better solution.

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
ccremer: yes, the same with LIEUTENANT_URL actually

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
ccremer: I have added `--kubeconfig "$(pwd)/../../../target/auth/kubeconfig"`, that right?

srueg: Depends on the directory you're currently in 🤷‍♂️

corvus-ch (approved):

srueg (approved):

PR #63, opened 2021-01-19
Document how to drop alert rules

corvus-ch:


docs/modules/ROOT/pages/how-tos/monitoring/remove_rules.adoc
```diff
@@ -0,0 +1,37 @@
+= Remove alert rules
+
+[abstract]
+Somtimes an alert rule doesn't makes sense.
```
tobru, line 4:
```suggestion
Sometimes an alert rule doesn't make any sense.
```

docs/modules/ROOT/pages/how-tos/monitoring/remove_rules.adoc
```diff
@@ -0,0 +1,37 @@
...
+Be it that the rule is build on assumptions that don't apply.
```
tobru, line 5:
```suggestion
Be it that the rule is built on assumptions that don't apply.
```

docs/modules/ROOT/pages/how-tos/monitoring/remove_rules.adoc
```diff
@@ -0,0 +1,37 @@
...
+[source,yaml]
+----
+parameters:
+  openshift4_monitoring:
+    alerts:
+      ignoreNames:
+        - MyAlertName
+----
++
+Do so on what ever hierarchy level is the most apropriate.
```
tobru, line 30:
```suggestion
Do so on whatever hierarchy level is the most appropriate.
```

tobru (changes requested):
Besides some typos LGTM. We'll see in practice if this is sufficient or need more explanation.

ccremer (approved):
I would have found the same typos :)

tobru (approved):

PR #73, opened 2021-03-10
Lookup TENANT_ID and REGION

corvus-ch:
Reduce the number of things one has to manually collect.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
@@ -73,10 +73,17 @@ export LDAP_PASSWORD="Your_LDAP_pw_here"
 .Access to various APIs
 [source,console]
 ----
-export CLOUDSCALE_TOKEN=<cloudscale-api-token> # From https://control.cloudscale.ch/user/api-tokens
-export GITLAB_TOKEN=<gitlab-api-token> # From https://git.vshn.net/profile/personal_access_tokens
-export COMMODORE_API_URL=<lieutenant-api-endpoint> # For example: https://api-int.syn.vshn.net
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -70,17 +70,27 @@ export LDAP_PASSWORD="Your_LDAP_pw_here"
 
 === Configure input
 
+.Access to various APIs
 [source,console]
 ----
 export CLOUDSCALE_TOKEN=<cloudscale-api-token> # From https://control.cloudscale.ch/user/api-tokens
 export GITLAB_TOKEN=<gitlab-api-token> # From https://git.vshn.net/profile/personal_access_tokens
-export REGION=rma # rma or lpg (without the zone number)
+export COMMODORE_API_URL=<lieutenant-api-endpoint> # For example: https://api-int.syn.vshn.net
```
simu, line 78: We probably want to add a note here that the `COMMODORE_API_URL` *must* have a trailing slash for the subsequent `curl` commands to work.

simu (approved):

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
srueg: I'd rather add the slash to the commands...

corvus-ch: I upted for adding the `/`. That was anyway my initial intention but somehow got lost.

simu: Then we **must** ensure that there's no trailing slash here, otherwise Lieutenant API isn't happy either:

```shell
$ curl https://api.syn.vshn.net/healthz
ok
$ curl https://api.syn.vshn.net//healthz
{"reason":"Path was not found"}
```

simu (approved):

PR #74, opened 2021-03-10
Add Dagobert secrets

corvus-ch:


docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -173,6 +173,25 @@ vault kv put clusters/kv/${TENANT_ID}/${CLUSTER_ID}/global-backup \
...
+
+  *)
+    echo "Can not determine copy source for the Dagobert secrets" >&2
+    export DAGOBERT_SOURCE=""
+  ;;
+esac
+
+# Copy the Dagobert OpenShift Node Collector Credentials
+vault kv get -format=json "${DAGOBERT_SOURCE}" | jq '.data.data' \
+  | vault kv put -cas=0 "clusters/kv/${TENANT_ID}/${CLUSTER_ID}/dagobert" -
```
simu, line 194: Maybe better if we only try to copy dagobert credentials if we actually found a source

```suggestion
if [ -n "${DAGOBERT_SOURCE}" ]; then
vault kv get -format=json "${DAGOBERT_SOURCE}" | jq '.data.data' \
  | vault kv put -cas=0 "clusters/kv/${TENANT_ID}/${CLUSTER_ID}/dagobert" -
fi
```

simu (approved):

docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
corvus-ch: I plaed it nice and added even more output.

simu (approved):

tobru:
As this is not cloudscale.ch specific, I'd say this belongs into a partial which can be reused for other installation documentation. And as this is really an ugly hack, copying from any random OCP4 cluster, I'd either point to the password manager, store the secret on another more generic location to copy from or at least make a note to SYN-356.

corvus-ch:
I disussed this with @simu and we agreed to do the follwoing:

* Place the Dagobert copy source in `clusters/kv/template/dagobert`
* Mention this workaround in SYN-356.

We also considered to go with `shared` or `global` instead of `template`. We opted for `template` as it is less likely to colide with what might come out of SYN-356.

For those not having access to SYN-356: It is a story in VSHNs internal ticket system and it is about the ability to have secrets defined on the global and the tenant level.

simu (approved):

srueg:
We can add a Vault policy for clusters to access `clusters/kv/data/global/*`: https://git.vshn.net/syn/synfra-apps/-/merge_requests/44

tobru (changes requested):
Please correct the typo in the commit message

tobru (approved):

PR #84, opened 2021-05-28
Remove installation steps regarding Ursula

ccremer:
The terraform module uses a public download location, so no need to provide our own download in the bootstrap bucket.


ccremer:
PR Draft status: Do not merge yet, Floaty isn't released yet and neither is https://github.com/appuio/terraform-openshift4-exoscale/pull/9

simu (approved):

PR #86, opened 2021-06-02
Update parameters learned from last setup

DebakelOrakel:
* Print out zones, when checking if imamge is available
* Fix url of image, when uploading to cloudscale
* Exchange $(whoami) for dedicated ENV, when initialising terraform backend
* Remove path to kubeconfig, as it is allready exported and non functional

Relates: VSHNOPS-3250


simu (approved):
LGTM

PR #102, opened 2021-08-03
Add runbook for adding and removing storage nodes

glrf:


glrf:
The last commit also touches the "replace a storage node" runbook. But it should mostly just switch to the new partials. The only semantic difference should be that I remove the unnecessary `--yes-i-really-mean-it` flag.

docs/modules/ROOT/partials/storage-ceph-remove-mon.adoc
```diff
@@ -0,0 +1,107 @@
...
+. Wait for operator to settle.
+Wait for a log message saying `done reconciling ceph cluster in namespace "syn-rook-ceph-cluster"`
++
+[source,bash]
+----
+kubectl --as=cluster-admin -n syn-rook-ceph-operator logs -f \
+  deploy/rook-ceph-operator
+----
+
+. Cordon node to replace and delete MON pod
```
simu, line 55: The text for this step should be configurable, currently the "Remove a node how-to" says "Cordon node to replace". Probably something like

```suggestion
. Cordon node to {node-operation} and delete MON pod
```

And set `:node-operation: replace` and `:node-operation: remove` accordingly in the how-tos.

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,243 @@
...
++
+[source,bash]
+----
+commodore catalog compile "${CLUSTER_ID}"
+----
+
+== Set alert silence
+
+:duration: +60 minutes
+include::partial$create-alertmanager-silence.adoc[]
```
simu, line 72: From what I can see, this whole section is identical between this how-to, adding a node and replacing a node. Probably makes sense to move this to a partial, as any changes here will need to be reflected in all three howtos anyway.

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,243 @@
...
+. Remove node ID and node resource for node that we want to remove from the Terraform state
++
+[source,bash]
+----
+terraform state rm "module.cluster.module.storage.random_id.node_id[$node_index]"
+terraform state rm "module.cluster.module.storage.exoscale_compute.nodes[$node_index]"
+----
+
+:delete-node: ${NODE_TO_REMOVE}
+include::partial$exoscale/delete-node.adoc[]
```
simu, line 225: This will completely break subsequent runs of `terraform apply` if any storage node except the one with the highest index in the Terraform state is removed.

I'd suggest extracting the node to remove from the Terraform state (highest ID in the state) and adjust the how-to to remove that node. Deleting the VM on Exoscale then becomes

```
terraform apply
```

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,243 @@
...
+  -l ceph.rook.io/pvc="${pvc_name}"
+kubectl --as=cluster-admin -n syn-rook-ceph-cluster delete pvc "${pvc_name}"
+----
+
+. Clean up PVC encryption secret
++
+[source,bash]
+----
+kubectl --as=cluster-admin -n syn-rook-ceph-cluster delete secret -l pvc_name="${pvc_name}"
+----
```
simu, line 179: Either add this to the `storage-ceph-remove-osd.adoc` partial (including scaling the operator step (conditional on another attribute) which is omitted here), or create another partial for this part, as it seems identical to the steps for replacing a node.

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,243 @@
...
+
+:delete-node: ${NODE_TO_REMOVE}
+include::partial$exoscale/delete-node.adoc[]
+
+. Clean up localstorage PV of decommissioned node
++
+[source,bash]
+----
+kubectl --as=cluster-admin delete pv "${pv_name}"
+----
```
simu, line 232: Pretty sure this can go into `delete-node.adoc`

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,243 @@
...
+
+. Remove node ID and node resource for node that we want to remove from the Terraform state
++
+[source,bash]
+----
+terraform state rm "module.cluster.module.storage.random_id.node_id[$node_index]"
+terraform state rm "module.cluster.module.storage.exoscale_compute.nodes[$node_index]"
+----
+
+:delete-node: ${NODE_TO_REMOVE}
```
simu, line 224: Defining the page attribute between list items breaks the list flow and the included items from the partial are numbered starting from 1. Generally, it's safer to define attributes at the top.

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,243 @@
...
+=== Remove old OSD
+. Make sure ArgoCD ran and reduced the target number of OSDs
++
+[source,bash]
+----
+kubectl --as=cluster-admin -n syn-rook-ceph-cluster \
+  get cephcluster cluster -o jsonpath='{.spec.storage.storageClassDeviceSets[0].count}'
+----
+
+:osd-node: ${NODE_TO_REMOVE}
```
simu, line 139: Define at top of the page

docs/modules/ROOT/pages/how-tos/exoscale/add_storage_node.adoc
```diff
@@ -0,0 +1,172 @@
...
++
+include::partial$get-hieradata-token-from-vault.adoc[]
++
+[source,bash]
+----
+export TF_VAR_lb_exoscale_api_user=$(vault kv get \
+  -format=json "clusters/kv/${TENANT_ID}/${CLUSTER_ID}/floaty" | jq '.data.data')
+export TF_VAR_lb_exoscale_api_key=$(echo "${TF_VAR_lb_exoscale_api_user}" | jq -r '.iam_key')
+export TF_VAR_lb_exoscale_api_secret=$(echo "${TF_VAR_lb_exoscale_api_user}" | jq -r '.iam_secret')
+----
```
simu, line 67: This is probably also required in `replace_storage_node.adoc`, not sure how I missed that.

docs/modules/ROOT/pages/how-tos/exoscale/add_storage_node.adoc
```diff
@@ -0,0 +1,172 @@
...
+# Find the new storage node
+kubectl --as=cluster-admin get nodes
+
+# Label the new storage node
+kubectl --as=cluster-admin label node <storage-XXX>\
+  node-role.kubernetes.io/storage=""
+
+kubectl --as=cluster-admin taint node -lnode-role.kubernetes.io/storage \
+  storagenode=True:NoSchedule
+----
```
simu, line 155: Why not just use the following as the current snippet isn't even copy-pasteable?

```suggestion
. Label and taint the new storage node
+
include partial$exoscale/label-taint-storage-nodes.adoc[]
```

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,243 @@
...
+----
+
+== Prepare Terraform environment
+
+include::partial$exoscale/configure-terraform-secrets.adoc[]
+
+include::partial$setup_terraform.adoc[]
+
+== Remove Node
+Make a note of the node you want to remove
```
simu, line 123: Maybe make this a list item?

docs/modules/ROOT/pages/how-tos/exoscale/replace_storage_node.adoc
```diff
@@ -180,54 +180,8 @@ kubectl --as=cluster-admin -n syn-rook-ceph-cluster get pods -w
...
-----
-
-. Check that the OSD is no longer listed in `ceph osd tree`
-+
-[source,bash]
-----
-kubectl --as=cluster-admin -n syn-rook-ceph-cluster exec -it deploy/rook-ceph-tools -- \
-  ceph osd tree
-----
+:osd-node: ${NODE_TO_REPLACE}
```
simu, line 183: Probably a good idea to move to the top of the page, even if it doesn't affect list numbering here.

docs/modules/ROOT/partials/storage-ceph-remove-mon.adoc
```diff
@@ -0,0 +1,107 @@
...
+----
+MON_ID=$(kubectl --as=cluster-admin -n syn-rook-ceph-cluster get pods \
+  -lapp=rook-ceph-mon -o wide \
+  | grep "{mon_node}" | cut -d- -f4)
+echo $MON_ID
+----
++
+TIP: You can skip the remaining steps in this section if `$MON_ID` is empty.
+
+ifeval::["{argocd_disabled}" != "true"]
```
simu, line 14: Might be better to use strings "yes" and "no" to make it clear that we're doing string compares

simu (changes requested):
There's quite a few sections that still appear to be replicated 1:1 (or quite close to 1:1) between the how-tos. I've marked the ones that I noticed. It would be nice to move those into partials so we don't have to maintain 2-3 copies of them (e.g. the whole prerequisites+local setup block).

docs/modules/ROOT/pages/how-tos/exoscale/add_storage_node.adoc
glrf: I suspected that using the approach from the partial could break, as we first label all worker nodes as storage nodes and then remove the storage label from the other ones.

That means for a short time all nodes will be labeled as storage nodes. If the rook operator ran at exactly that time, couldn't it schedule the new OSD on a wrong node?

simu: Maybe for other clouds where we can use CSI-provisioned PVCs, but on Exoscale, no new OSD will be scheduled unless there's an available localstorage PV and the localstorage PV constrains the pod to a storage node anyway.

glrf:
This adds a small ordering change in the replace runbook to get cleaner partials. But AFAIK this should still work in that order.

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,140 @@
...
+  ${CLUSTER_ID}.yml
+
+yq eval -i ".parameters.rook_ceph.ceph_cluster.node_count -= 1" \
+  ${CLUSTER_ID}.yml
+----
++
+[NOTE]
+====
+Ceph can't scale below 3 storage nodes, which is the default number of nodes.
+Please check that this update doesn't reduce the number of storage nodes below 3, before continuing.
```
simu, line 55:
```suggestion
Please ensure that this update doesn't reduce the number of storage nodes to less than 3 before continuing.
```

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,140 @@
...
+yq eval -i ".parameters.openshift4_terraform.terraform_variables.storage_count -= 1" \
+  ${CLUSTER_ID}.yml
+
+yq eval -i ".parameters.rook_ceph.ceph_cluster.node_count -= 1" \
+  ${CLUSTER_ID}.yml
+----
++
+[NOTE]
+====
+Ceph can't scale below 3 storage nodes, which is the default number of nodes.
```
simu, line 54:
```suggestion
Ceph can't scale to less than 3 storage nodes, which is the default number of nodes.
```

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -0,0 +1,140 @@
...
+----
+# Grab JSON copy of current Terraform state
+terraform state pull > .tfstate.json
+
+node_count=$(jq  -r \
+  '.resources[] |
+   select(.module=="module.cluster.module.storage" and .type=="exoscale_compute") |
+   .instances | length' \
+   .tfstate.json)
+# Make sure the number of nodes is conistent with what we configured earlier
```
simu, line 98:
```suggestion
# Verify that the number of nodes is one more than we configured earlier.
```

docs/modules/ROOT/partials/storage-ceph-remove-mon.adoc
```diff
@@ -0,0 +1,107 @@
...
+----
+MON_ID=$(kubectl --as=cluster-admin -n syn-rook-ceph-cluster get pods \
+  -lapp=rook-ceph-mon -o wide \
+  | grep "{mon-node}" | cut -d- -f4)
+echo $MON_ID
+----
++
+TIP: You can skip the remaining steps in this section if `$MON_ID` is empty.
+
+ifeval::["{mon-argocd-disabled}" != "yes"]
```
simu, line 14: shouldn't this be
 
```suggestion
ifeval::["{mon-argocd-disabled}" == "yes"]
```

docs/modules/ROOT/partials/storage-ceph-remove-mon.adoc
```diff
@@ -0,0 +1,107 @@
...
+  -p '[{
+    "op": "replace",
+    "path": "/spec/healthCheck/daemonHealth/mon",
+    "value": {}
+  }]'
+kubectl --as=cluster-admin -n syn-rook-ceph-operator delete pods \
+  -l app=rook-ceph-operator
+----
+
+ifeval::["{mon-argocd-disabled}" != "yes"]
```
simu, line 104: same as above

simu (approved):
LGTM now. I've left some wording suggestions inline.

Might be good to test the new how-tos once more (if you haven't yet)

PR #104, opened 2021-08-05
Bump OCP to 4.8

mhutter:
Signed-off-by: Manuel Hutter <manuel@hutter.io>


corvus-ch:
The same aplies here as I wrote at https://github.com/appuio/terraform-openshift4-cloudscale/issues/27#issuecomment-893382883_

simu:
Install docs were updated in #114 and #110

PR #107, opened 2021-08-12
Fix Exoscale storage node management how-tos

simu:
* Remove node from crushmap instead of OSD id
* Fix how we use the `delete-pvs` page attribute
* Fix copy-paste errors in "Remove storage node" how-to
* Scale down the rook-ceph operator in "Remove storage node" how-to

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
+++ b/docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
@@ -7,6 +7,7 @@
 :osd-replace-list: $NODE_TO_REMOVE
 
 :mon-operation: remove
+:mon-argocd-autosync-already-disabled: yes
 :mon-replace-list: $NODE_TO_REMOVE
 
 :node-delete-list: ${NODE_TO_REMOVE}
--- a/docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
+++ b/docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
@@ -11,7 +11,7 @@
 
 :node-delete-list: ${NODE_TO_REMOVE}
 :delete-nodes-manually: no
-:delete-pvs: $old_pv_names
+:delete-pvs: old_pv_names
 
 [abstract]
 --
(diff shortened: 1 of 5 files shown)


docs/modules/ROOT/partials/storage-ceph-remove-osd.adoc
```diff
@@ -33,7 +33,7 @@ for node in $(echo -n {osd-replace-list}); do
   kubectl --as=cluster-admin -n syn-rook-ceph-cluster exec -it deploy/rook-ceph-tools -- \
     ceph osd purge "${osd_id}"
   kubectl --as=cluster-admin -n syn-rook-ceph-cluster exec -it deploy/rook-ceph-tools -- \
-    ceph osd crush remove "${osd_id}"
+    ceph osd crush remove "${node}"
```
bastjan, line 36: Does this command really use the node id when all others use the osd id? Maybe the command fails because purge already removes the osd from the crushmap?

docs/modules/ROOT/partials/exoscale/delete-node.adoc
```diff
@@ -54,7 +56,7 @@ endif::[]
 +
 [source,bash,subs="attributes+"]
 ----
-for pv_name in $(echo -n {delete-pvs}); do
+for pv_name in $(echo -n ${delete-pvs}); do
```
bastjan, line 59: In (all) other templates the `$` is part of the attribute `:mon-replace-list: $NODE_TO_REMOVE`. Should we decide on one style?

bastjan: never mind then 🤦‍♀️

docs/modules/ROOT/partials/storage-ceph-remove-osd.adoc
simu: Yes, the OSD is removed from the crush map already with `osd purge`, but the node still remains. With `crush remove` you can remove a failure domain (node for us) from the crush map.

bastjan (approved):

docs/modules/ROOT/pages/how-tos/exoscale/remove_storage_node.adoc
```diff
@@ -119,6 +119,18 @@ kubectl --as=cluster-admin -n syn-rook-ceph-cluster \
   get cephcluster cluster -o jsonpath='{.spec.storage.storageClassDeviceSets[0].count}'
 ----
 
+. Disable ArgoCD auto sync for component `rook-ceph`
++
+include::partial$disable-argocd-autosync.adoc[]
+
+. Scale down the Rook-Ceph operator
```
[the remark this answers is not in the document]
simu: good point, we need to scale it up, before potentially moving MONs, as that won't work if the operator is scaled down.

glrf (approved):

PR #110, opened 2021-08-25
Update exoscale install doc to openshift 4.8

glrf:


docs/modules/ROOT/pages/how-tos/exoscale/install.adoc
```diff
@@ -193,6 +195,7 @@ git push
 
 popd
 ----
+<1> Can be removed as soon as Project Syn is kuberenetes version aware and we can set this automatically.
```
tobru, line 198:
```suggestion
<1> Can be removed as soon as Project Syn is Kubernetes version aware and we can set this automatically.
```

glrf: I guess I still need more coffee..

tobru: :coffee:

PR #119, opened 2021-09-06
Add documentation on worker groups and volume sizes

glrf:


tobru:
Parts of this PR seem to copy content from https://products.docs.vshn.ch/products/appuio/managed/ocp4_cloudscale.html#_default_configuration_minimum_requirements. How should we avoid having the same information and different places? We could remove it from the product docs and link here, or vice-versa...

docs/modules/ROOT/pages/references/cloudscale/config.adoc
```diff
@@ -0,0 +1,132 @@
...
+The OpenShift cluster setup can be customized by configuring `.parameters.openshift4_terraform.terraform_variables`.
+This page gives an overview of the configuration options, when setting up a cluster on Cloudscale.
+Futher information on the terraform module and its configuration can be found on https://github.com/appuio/terraform-openshift4-cloudscale[GitHub].
+
+
+
+== Nodes
+
+Through Terraform variables you have the ability to control the instance size, count, and partition size for each VM type, as well as the option to configure additional worker groups. 
+
```
simu, line 13:
```suggestion

See the https://www.cloudscale.ch/en/pricing[cloudscale.ch pricing doucumentation] for available VM flavors.
```

simu (changes requested):
Some typos (might be worth to have a partial for the prelude of the two configuration pages with an attribute to inject `cloudscale.ch` and `Exoscale` as appropriate), and a suggestion to add warnings for configurations which currently aren't supported (although those duplicate more information from the product documentation)

glrf:
I actually did not know of the product documentation page. I agree there is some overlap but in the end the two pages serve two different purposes. I'm not sure how useful it actually is to deduplicate this.

As for using partials to unify the two configuration page I'm a little torn. Yes most of it is very similar, but there are differences in naming all over the place. So we would need to use a lot of templating. So the main question is: do we want to maintain two very similar but simple pages, or do we want a single complicated page.

simu (changes requested):
The remaining suggestions for the Exoscale storage nodes still need to be incorporated.

Re partials: I suggested them because I found the exact same typos in the preamble of both documents.

corvus-ch (approved):

simu (approved):

tobru:
> I actually did not know of the product documentation page. I agree there is some overlap but in the end the two pages serve two different purposes. I'm not sure how useful it actually is to deduplicate this.

Let's keep it as you suggested in this PR, I think that's fine for now.

tobru (approved):

PR #125, opened 2021-09-29
Notes and missing steps for cloudscale install instructions

haasad:
- Additional steps when using a custom apps domain
- Add missing step for etcd DNS records
- Add note about image upload to lpg

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
@@ -305,7 +305,7 @@ module "cluster" {
   additional_worker_groups = {}
 }
 EOF
-terraform apply -target "module.cluster.local_file.lb_hieradata[0]"
+terraform apply -target "module.cluster.module.lb.module.hiera.local_file.lb_hieradata"
 ----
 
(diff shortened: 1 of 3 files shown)


simu (approved):
LGTM. We might want to consider making the zone for the image upload a variable, so the command itself remains copy-pasteable. Something like

```. Import the image to Cloudscale
+
[source,bash,subs="attributes+"]
----
zone=rma1
curl -i -H "$AUTH_HEADER" \
  -F url="$(mc share download --json "$\{CLUSTER_ID\}/$\{CLUSTER_ID\}-bootstrap-ignition/rhcos-{ocp-minor-version}.raw" | jq -r .url)" \
  -F name='RHCOS {ocp-minor-version}' \
  -F zones="${zone}" \
  -F slug=rhcos-{ocp-minor-version} \
  -F source_format=raw \
  -F user_data_handling=pass-through \
  https://api.cloudscale.ch/v1/custom-images/import
----
```

ccremer (approved):

haasad:
Another option is to use something like `-F zones="${REGION}1", as the REGION variable is exported earlier in the setup.
However its also possible to upload an image to both zones by using:
```
...
-F zones=rma1
-F zones=lgp1
...
```

simu:
I didn't see that we already set `${REGION}` appropriately. I think it's fine to use that variable then. IIRC we don't upload to both zones to minimize cost overhead.

PR #135, opened 2021-11-18
Fix local_storage_operator manifest

bastjan:


glrf (approved):
I'm surprised we never noticed that..

ccremer (approved):

PR #136, opened 2021-11-18
Fix DNS NS record creation on Exoscale

bastjan:
The DNS records seem to have a implicit dependency on the hieradata.


simu (approved):
It's possible that the DNS output isn't printed because the other output depends on the LB hieradata (so the MR URL can be printed).

PR #139, opened 2021-11-22
Fix Terraform command for creating hieradata in clouscale.ch install guide

simu:
The old command confused Terraform: the command resulted in the correct plan but the subsequent apply ended with

```
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

This commit adjusts the command to only target the whole vshn-lbaas-hieradata module resource, which should have the same
outcome, namely that the LB hieradata MR and resources it depends on are created.


bastjan (approved):
Interesting, pretty sure it worked at one point 🤷‍♀️

rxbn (approved):
LGTM

PR #142, opened 2021-11-30
Update documentation to add optional Cilium steps

bastjan:
This PR adds additional steps to install Cilium using the provided component-cilium.

### DIFF
--- a/docs/modules/ROOT/partials/install/prepare-installer.adoc
+++ b/docs/modules/ROOT/partials/install/prepare-installer.adoc
@@ -1,5 +1,3 @@
-For the following steps, change into a clean directory (for example a directory in your home).
-
 [CAUTION]
 These are the only steps which aren't idempotent and have to be completed uninterrupted in one go.
 If you have to recreate the install config or any of the generated manifests you need to rerun all of the subsequent steps.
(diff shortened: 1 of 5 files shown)


docs/modules/ROOT/partials/install/prepare-terraform.adoc
```diff
@@ -1,17 +1,3 @@
-
```
simu, line 1: This needs to stay in the partial, or the Exoscale install instructions need to be updated accordingly.

docs/modules/ROOT/partials/install/prepare-installer.adoc
```diff
@@ -1,5 +1,3 @@
-For the following steps, change into a clean directory (for example a directory in your home).
```
simu, line 1: This needs to stay in the partial or the Exoscale instructions need to be updated accordingly.

docs/modules/ROOT/partials/install/prepare-installer.adoc
```diff
@@ -74,6 +82,16 @@ openshift-install --dir "${INSTALLER_DIR}" \
...
   spec.domain apps.example.com
 ----
++
+. Cilium Optional: Copy pre-rendered Cilium manifests
++
+[%collapsible]
+====
+[source,bash]
+----
+cp catalog/manifests/cilium/olm/* target/manifests/
```
simu, line 92: This will show up in the Exoscale install instructions, but the catalog won't be present, because the early catalog rendering is only part of the cloudscale.ch install instructions.

simu (commented):
Looks good for cloudscale.ch, but probably breaks the Exoscale install how-to.

docs/modules/ROOT/partials/install/prepare-installer.adoc
bastjan: 🤦‍♀️ forgot to move the commodore init into a partial

simu (approved):

PR #147, opened 2021-12-06
Add missing steps to decommissioning documentation

rxbn:
- Fix small typos
- Add missings steps

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/decommission.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/decommission.adoc
@@ -25,7 +25,7 @@ Steps to remove an OpenShift 4 cluster from https://cloudscale.ch[cloudscale.ch]
 +
 [source,bash]
 ----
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/cloudscale/decommission.adoc
```diff
@@ -127,6 +127,12 @@ Go to https://control.vshn.net/syn/lieutenantclusters
 +
 - Delete cluster entry using the delete button
 
+. Delete all volumes in cloudscale.ch project
```
simu, line 130:
```suggestion
. Delete all remaining volumes which were associated with the cluster in the cloudscale.ch project.
+
TIP: This step is required because the csi-cloudscale driver doesn't have time to properly cleanup PVs when the cluster is decommissioned with `terraform destroy`.
```

simu (approved):
LGTM overall, some suggestions inline

PR #150, opened 2021-12-15
Add how-to for `oc debug node`

simu:


ccremer (approved):
LGTM. Do we want/need to mention that this only works as long as there's a responsive Kubernetes API?

corvus-ch (approved):

PR #151, opened 2021-12-21
How-To: Increase Storage Size on an OCP Logging Elasticsearch instance

bastjan:
This PR describes how to increase the underlying storage size of the OpenShift Cluster Logging Elasticsearch instance.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
+++ b/docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
@@ -24,35 +24,7 @@ During the resize, the Elasticsearch cluster will experience reduced performance
 
 == Prepare local environment
 
-. Configure API access
-+
-[source,bash]
-----
-export COMMODORE_API_URL=https://api.syn.vshn.net <1>
-export COMMODORE_API_TOKEN=<your lieutenant token>
-
-# Set Project Syn cluster and tenant ID
-export CLUSTER_ID=<lieutenant-cluster-id> # Looks like: c-cluster-id-1234
-export TENANT_ID=$(curl -sH "Authorization: Bearer ${COMMODORE_API_TOKEN}" ${COMMODORE_API_URL}/clusters/${CLUSTER_ID} | jq -r .tenant)
-----
-<1> Replace with the API URL of the desired Lieutenant instance.
-
-. Create a local directory to work in and compile the cluster catalog
-+
-[source,bash]
-----
-export WORK_DIR=/path/to/work/dir
-mkdir -p "${WORK_DIR}"
-pushd "${WORK_DIR}"
-
-commodore catalog compile "${CLUSTER_ID}"
-----
-+
-[TIP]
-====
-We strongly recommend creating an empty directory, unless you already have a work directory for the cluster you're about to work on.
-This guide will run Commodore in the directory created in this step.
-====
+include::partial$commodore-init.adoc[]
 
 == Increase PVC sizes
 
--- a/docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
+++ b/docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
@@ -1,0 +1,283 @@
+= Increase Storage Size of an OCP Logging Elasticsearch instance
+
+[abstract]
+This page describes how to increase the underlying storage size of the OpenShift Cluster Logging Elasticsearch instance.
+
+== Starting situation
+
+* You already have an OpenShift 4 with Cluster Logging enabled.
+* The Cluster Logging instance is managed and is of type `elasticsearch`.
+* You have admin-level access to the cluster.
+* You want to increase the storage size of the Elasticsearch cluster.
+
+== Prerequisites
+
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
```diff
@@ -0,0 +1,283 @@
...
+pushd "${WORK_DIR}"
+
+commodore catalog compile "${CLUSTER_ID}"
+----
++
+[TIP]
+====
+We strongly recommend creating an empty directory, unless you already have a work directory for the cluster you're about to work on.
+This guide will run Commodore in the directory created in this step.
+====
```
ccremer, line 51: I think here we can just link to https://syn.tools/syn/how-tos/compile-catalog.html

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
```diff
@@ -0,0 +1,283 @@
...
+git push
+popd
+----
++
+. Compile catalog
++
+[source,bash]
+----
+commodore catalog compile ${CLUSTER_ID} --push -i
+----
```
ccremer, line 84: Maybe this can also be combined with https://syn.tools/syn/how-tos/change-parameter.html

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
```diff
@@ -0,0 +1,283 @@
...
+  --as=cluster-admin \
+  -n openshift-logging \
+  exec "${es_pods[1]}" \
+  -c elasticsearch \
+  -- es_util --query="_cluster/settings" -XPUT -d '{ "persistent": { "cluster.routing.allocation.enable" : "primaries" } }'
+----
++
+Example output:
++
+[source,json]
```
ccremer, line 178: Maybe just give the box a title? (applies to other locations as well)
```suggestion
.Example output
[source,json]
```

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
```diff
@@ -0,0 +1,283 @@
...
+  -n openshift-logging \
+  get pods \
+  --watch
+----
+<1> Replace with deployment name found in previous step.
++
+.. Wait until cluster becomes healthy again.
++
+[WARNING]
+Make sure the status is `green` or `yellow` before proceeding.
```
ccremer, line 242: I think we should decide on which "color", this is unclear
```suggestion
Make sure the status is at least `yellow` before proceeding.
```

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
```diff
@@ -0,0 +1,283 @@
+= Increase Storage Size of an OCP Logging Elasticsearch instance
+
+[abstract]
+This page describes how to increase the underlying storage size of the OpenShift Cluster Logging Elasticsearch instance.
+
```
ccremer, line 5: I think we should mention the impact of the resize.
```suggestion

[IMPORTANT]
This is a disruptive operation!
During the resize, the Elasticsearch cluster will experience downtime for a few minutes to hours.

```

bastjan: There is no downtime since we have single redundancy!

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
bastjan: I'm not a fan. In my experience it's way easier to follow a guide on one page and not have to jump between 2-3 pages.

It would be a bit inconsistent too. All other mentions in the docs show how to compile commodore directly on the page and don't link to syn.tools.

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
bastjan: There will be reduced performance for a few minutes though and new logs are delayed. I try to write it like this.

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
bastjan: I thing it's clearer if both possible colors are explicitly mentioned. For me the cluster stayed green while restarting two of three nodes.

docs/modules/ROOT/pages/how-tos/logging/increase-elasticsearch-storage-size.adoc
ccremer: Ok, but at least make a partial, otherwise this is unmaintainable.

bastjan: 👍

PR #153, opened 2021-12-27
Document how to change vSphere parameters properly

mhutter:


docs/modules/ROOT/pages/how-tos/vsphere/change-vsphere-params.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+[abstract]
+This checklist should help in changing the parameters in all places necessary.
+
+Example scenario: The Datacenter was renamed in vSphere.
+
+== Checlist
+
+[options="interactive"]
+* [ ] Update the `vsphere-config-secret` in the `kube-suystem` namespace.
```
corvus-ch, line 11:
```suggestion
* [ ] Update the `vsphere-config-secret` in the `kube-system` namespace.
```

corvus-ch (commented):
This is quit fuzzy to me. I do not exactly get when this gets triggered. Further, the steps hints where to change chings but don't tell you what needs to be chagned.

This might be, because I am not familiar with vSphere.

mhutter:
> I do not exactly get when this gets triggered

See ASPC-395. There, it got only updated in some place, triggering various other issues in response. Hence this documentation.

corvus-ch (approved):

PR #154, opened 2021-12-29
How to deal with failing node drains during maintenance

haasad:
- Make it more explicit that maintenance is only finished once all nodes are upgraded
- Add troubleshooting info for the situation where the Machine Config operator fails to drain nodes during the upgrade procedure

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
+++ b/docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
@@ -208,7 +208,7 @@ kubectl --as=cluster-admin -n openshift-machine-config-operator \
 oc --as=cluster-admin adm drain <node-name> --delete-emptydir-data --ignore-daemonsets --force --grace-period=0
 ----
 +
-If manually force-draining the node isn't successfull, check which pods are still running on the node with `oc describe node <node-name>` and force delete the pods.
+If manually force-draining the node isn't successful, check which pods are still running on the node with `oc describe node <node-name>` or `oc get pods --all-namespaces --field-selector spec.nodeName=<node-name>` and force delete any non-daemonset pods shown in the output.
 The Machine Config operator should then be able to continue with the node upgrades.
 Depending on what's blocking the drain, these steps may have to be repeated for several nodes.
 
--- a/docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
+++ b/docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
@@ -208,7 +208,9 @@ kubectl --as=cluster-admin -n openshift-machine-config-operator \
 oc --as=cluster-admin adm drain <node-name> --delete-emptydir-data --ignore-daemonsets --force --grace-period=0
 ----
 +
-If manually force-draining the node isn't successfull, check which pods are still running on the node with `oc describe node <node-name>` and force delete the pods. The Machine Config operator should then be able to continue with the node upgrades. Depending on what's blocking the drain, these steps may have to be repeated for several nodes.
+If manually force-draining the node isn't successfull, check which pods are still running on the node with `oc describe node <node-name>` and force delete the pods.
+The Machine Config operator should then be able to continue with the node upgrades.
+Depending on what's blocking the drain, these steps may have to be repeated for several nodes.
 
 * If nodes get stuck in `NotReady` during the upgrade process, check whether the VM got stuck trying to reboot itself into the new image:
 . Login to the cloud provider's web console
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
```diff
@@ -192,6 +194,15 @@ kubectl --as=cluster-admin -n openshift-machine-config-operator \
...
 ----
 
+* If the Machine Config operator fails to drain a node, you may have to force-drain the node:
++
+[source,console]
+----
+oc adm drain <node-name> --delete-emptydir-data --ignore-daemonsets --force --grace-period=0
+----
++
+If manually force-draining the node isn't successfull, check which pods are still running on the node with `oc describe node <node-name>` and force delete the pods. The Machine Config operator should then be able to continue with the node upgrades. Depending on what's blocking the drain, these steps may have to be repeated for several nodes.
```
simu, line 204: One sentence per line.

I'd propose to simply use `oc --as=cluster-admin adm drain --ignore-daemonsets --delete-emptydir-data --force --grace-period=0` to force drain nodes instead of manually finding stuck pods and force deleting them.

docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
```diff
@@ -86,6 +86,8 @@ oc --as=cluster-admin get mcp -w
 ----
 ====
 
+. The maintenance of the cluster is only finished once all nodes, including all worker nodes have been upgraded and all `MachineConfigPools` show `Updated=True`. Never leave the cluster in a state with pending node upgrades. If the Machine Config operator can't drain a node (eg. because of violating PodDisruptionBudgets) you may have to manually force-drain a node or even manually delete pods that block the node drain. Always open a follow-up ticket to investigate the underlying issues if manual intervention is required.
```
simu, line 89: Please keep documentation sources to one sentence per line.

Also, I'd move this section to the `IMPORTANT` box just above it.

docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
haasad: From experience: If a PDB prevents a pod from being deleted, the drain won't work even with `--force`, only manually deleting the pod allows the operator to continue

simu: Ah, good point, I forgot about that.

docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
haasad: In essence, this was already covered in the box above imho. I just want to make it extra obvious what's required for the maintenance to be considered finished, because we had several cases of pending node upgrades after the maintenance in the last weeks.

docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
```diff
@@ -86,6 +86,15 @@ oc --as=cluster-admin get mcp -w
 ----
 ====
 
+. The maintenance of the cluster is only finished once all nodes, including all worker nodes have been upgraded and all `MachineConfigPools` show `Updated=True`.
++
+[IMPORTANT]
+====
+Never leave the cluster in a state with pending node upgrades.
+If the Machine Config operator can't drain a node (eg. because of violating PodDisruptionBudgets) you may have to manually force-drain a node or even manually delete pods that block the node drain.
```
simu, line 94: I'm surprised Vale didn't complain about the `eg.` but that's possibly because that should be spelled `e.g.`.

```suggestion
If the Machine Config operator can't drain a node (for example because doing so would violate a `PodDisruptionBudget`) you may have to manually force-drain a node or even manually delete pods that block the node drain.
```

docs/modules/ROOT/pages/how-tos/update_maintenance.adoc
```diff
@@ -192,6 +201,17 @@ kubectl --as=cluster-admin -n openshift-machine-config-operator \
...
 ----
 
+* If the Machine Config operator fails to drain a node, you may have to force-drain the node:
++
+[source,console]
+----
+oc --as=cluster-admin adm drain <node-name> --delete-emptydir-data --ignore-daemonsets --force --grace-period=0
+----
++
+If manually force-draining the node isn't successfull, check which pods are still running on the node with `oc describe node <node-name>` and force delete the pods.
```
simu, line 211:
```suggestion
If manually force-draining the node isn't successful, check which pods are still running on the node with `oc describe node <node-name>` or `oc get pods --all-namespaces --field-selector spec.nodeName=<node-name>` and force delete any non-daemonset pods shown in the output.
```

simu (approved):

PR #156, opened 2022-01-06
Document operator deletion and troubleshooting

anothertobi:


bastjan (commented):
Looks good overall. Some command-line samples could help for the troubleshooting steps IMO.

corvus-ch (changes requested):
When the operator was managed with Syn, removing it from the cluster catalog, all related objects should be removed. At least that is my expectation. @simu Am I missing something? So this needs either a bug report or the text needs clarification (manual removal only needed manually created or in the troubleshooting case).

bastjan:
> When the operator was managed with Syn, removing it from the cluster catalog, all related objects should be removed.

The RedHat operator removal is super, super buggy from my experience. Some objects get orphaned. Pretty sure they messed up all owner-references.

Maybe we need to engineer a solution for that issue and find the orphans?

simu:
> 

Yeah, that's been my experience too, Project Syn can't clean up properly because there's owner references missing on some of the objects created by OLM.

Might be worth creating a support case to see what RH have to say for themselves (although I suspect that's going to end up going in circles).

anothertobi:
Uninstalling an operator will not clean up all things [1]:

> This action does not remove resources managed by the Operator, including custom resource definitions (CRDs) and custom resources (CRs). Dashboards and navigation items enabled by the web console and off-cluster resources that continue to run might need manual clean up. To remove these after uninstalling the Operator, you might need to manually delete the Operator CRDs.

I will add some `kubectl` commands to the troubleshooting section.

[1] https://docs.openshift.com/container-platform/4.9/operators/admin/olm-deleting-operators-from-cluster.html#olm-deleting-operators-from-a-cluster-using-web-console_olm-deleting-operators-from-a-cluster

corvus-ch (approved):
Not sure if this guide stands its value when used, but improvements can be made when issues are found.

PR #161, opened 2022-02-22
Use kubectl patch to enable proxy protocol on the ingress controller

haasad:
Proxy protocol on the cluster is only enabled once the cluster is synthesized. However the installation process gets stuck after bootstrapping the infra nodes, because ingress doesn't work (LBs already use proxy protocol).

This PR adds a manual step with kubectl patch that enables the proxy protocol on the ingress controller so the installation process can resume.


haasad:
@glrf Perfect timing :see_no_evil: thanks for fixing

PR #163, opened 2022-03-11
Adjust Exoscale install how-to to provide all steps for Cilium

simu:
We've tested the installation on Exoscale with the latest component-cilium (which will be released soon).


bastjan (approved):
🎉

PR #166, opened 2022-03-15
Add step to remove initial CA data from kubeconfig

simu:
This needs to be done after the production API certificate is deployed by Project Syn (either Let's Encrypt, or a bought certificate).

Also move the step to save the admin kubeconfig in the password manager after the new step, so we don't need to upload the kubeconfig twice.

## TODO

- [x] Rebase after #165 is merged

### DIFF
--- a/docs/modules/ROOT/partials/install/finalize_part1.adoc
+++ b/docs/modules/ROOT/partials/install/finalize_part1.adoc
@@ -1,11 +1,4 @@
-. Save the admin credentials in the https://cloud.passbolt.com/vshn[password manager].
-You can find the password in the file `target/auth/kubeadmin-password` and the kubeconfig in `target/auth/kubeconfig`
-+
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/partials/install/finalize_part1.adoc
```diff
@@ -1,11 +1,4 @@
-. Save the admin credentials in the https://cloud.passbolt.com/vshn[password manager].
-You can find the password in the file `target/auth/kubeadmin-password` and the kubeconfig in `target/auth/kubeconfig`
-+
-[source,bash]
-----
-popd
-ls -l ${INSTALLER_DIR}/auth/
-----
+:dummy:
```
corvus-ch, line 1: What is this doing?

corvus-ch (approved):

PR #168, opened 2022-03-21
Add minor upgrade checklist

corvus-ch:


docs/modules/ROOT/pages/how-tos/new_minor.adoc
```diff
@@ -0,0 +1,46 @@
...
+* API versions of resource types used
+* Alert rules
++
+The main suspect is `openshift4-monitoring`.
+But also lookout for components that define alert rules but are no longer picked up by cluster monitoring.
+
+. Conduct install on all supported cloud providers
++
+Use the installation documentation to install a new cluster.
+Do so for each supported hosting provider.
```
corvus-ch, line 45: Where is this defined?

simu (approved):
LGTM. Instead of leaving a pile of English suggestions, I directly pushed a commit which primarily adjusts wording.

docs/modules/ROOT/pages/how-tos/new_minor.adoc
```diff
@@ -0,0 +1,50 @@
+= Checklist to prepare for a new minor version of OpenShift 4
+
+. Read and summarize release notes
++
+Read the release notes.
+Look for things that affect how we use OpenShift.
+Ensure each change which impacts us gets addressed.
++
+Write a summary and add it to xref:oc4:ROOT:references/release_notes.adoc[our release notes summary].
```
tobru, line 9: And remove OpenShift versions which we don't support anymore (usually current version - 1, but I suggest keeping current version - 2)

docs/modules/ROOT/pages/how-tos/new_minor.adoc
```diff
@@ -0,0 +1,50 @@
...
+====
++
+[TIP]
+====
+Instead of using an existing cluster, you can also set up one with version `4.<n-1>` and then upgrade this one.
+====
+
+. Check compatibility of components
++
+Check if all core Commodore components are compatible with the new version.
```
tobru, line 31: What are the "core Commodore components"? Where is this defined?

docs/modules/ROOT/pages/how-tos/new_minor.adoc
```diff
@@ -0,0 +1,50 @@
...
+Address all issues as necessary.
+If it's straight-forward, directly create a fix.
+Otherwise, create a follow-up story to address the issue.
++
+The best way to find an address issues is to inspect an upgraded and or newly installed cluster using the new minor version.
```
tobru, line 36:
```suggestion
The best way to find and address issues is to inspect an upgraded and or newly installed cluster using the new minor version.
```

docs/modules/ROOT/pages/how-tos/new_minor.adoc
```diff
@@ -0,0 +1,50 @@
...
+Things to look for:
++
+* API versions of resource types used
+* Alert rules
+* Logic which generates configuration based on the cluster's OpenShift version.
++
+One component which definitely needs to be updated for a new minor version is `openshift4-monitoring`.
```
tobru, line 43:
```suggestion
One component which definitely needs to be updated for a new minor version is https://github.com/appuio/component-openshift4-monitoring[`openshift4-monitoring`].
```

docs/modules/ROOT/pages/how-tos/new_minor.adoc
tobru: IMHO here: https://products.docs.vshn.ch/products/appuio/managed/ocp4.html (yes, this list needs to be updated)

tobru (approved):
A few minor things, and then we're good to go. Thanks.

simu (approved):

docs/modules/ROOT/pages/how-tos/new_minor.adoc
```diff
@@ -0,0 +1,52 @@
...
+====
++
+[TIP]
+====
+Instead of using an existing cluster, you can also set up one with version `4.<n-1>` and then upgrade this one.
+====
+
+. Check compatibility of components
++
+Check if all core Commodore components{fn-core-components} are compatible with the new version.
```
tobru, line 33:
```suggestion
Check if all core Commodore components footnote:[Core Commodore components are those included by the distribution `openshift4` and its parent hierarchy levels.] are compatible with the new version.
```

There is no need for the complicated `{fn-core-components}` construct here...

tobru (approved):

docs/modules/ROOT/pages/how-tos/new_minor.adoc
corvus-ch: I kind of disliked the interleaved sentences when looking at the asciidoc. It does not matter when rendered.

PR #184, opened 2022-06-14
Add how-to for migrating from LDAP to Keycloak authentication

simu:


simu:
no longer relevant, closing

PR #187, opened 2022-07-14
Update Cloudscale flavors

bastjan:


docs/modules/ROOT/pages/references/cloudscale/config.adoc
```diff
@@ -148,11 +148,11 @@ parameters:
     terraform_variables:
       root_disk_size: 128
       worker_count: 4
-      worker_flavor: "plus-4"
+      worker_flavor: "plus-4-2"
```
simu, line 151: This is a bad example as we state just above that we don't support worker flavors < plus-16-4.

simu (approved):

PR #188, opened 2022-07-19
Add missing `--as=cluster-admin` in cloudscale.ch "Add node" how-to

simu:


schemen:
LGTM

docs/modules/ROOT/pages/how-tos/cloudscale/add_node.adoc
```diff
@@ -58,12 +59,12 @@ include::partial$install/approve-node-csrs.adoc[]
 +
 [source,bash]
 ----
-kubectl get node
+kubectl --as=cluster-admin get node
```
glrf, line 62: I think technically we don't need cluster-admin anymore to get nodes. But doesn't hurt either and might not be true on all clusters, so :shrug:

glrf (approved):

PR #192, opened 2022-07-22
Use qcow2 format for custom image import and delete bootstrap bucket

haasad:
cloudscale.ch now supports custom image import in qcow2 format. There's no need anymore to convert the image to `raw` and you only need to upload a ~2.5G file to S3 instead of 16G for the raw image.

Also add steps to delete the file from the bucket after the cutom image is successfully imported

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
@@ -45,7 +45,6 @@ It's currently very specific to VSHN and needs further changes to be more generi
 * `yq` https://mikefarah.gitbook.io/yq[yq YAML processor] (version 4 or higher)
 * `openshift-install` (direct download: https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-{ocp-minor-version}/openshift-install-linux.tar.gz[linux], https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-{ocp-minor-version}/openshift-install-mac.tar.gz[macOS])
 * `vault` https://www.vaultproject.io/docs/commands[Vault CLI]
-* `qemu-img`
 * `curl`
 * `gzip`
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
```diff
@@ -175,15 +173,32 @@ The output should be
...
+  -F source_format=qcow2 \
   -F user_data_handling=pass-through \
   https://api.cloudscale.ch/v1/custom-images/import
 ----
 
+. Wait for custom image import to complete
++
+[source,bash,subs="attributes+"]
+----
+while [ "$(curl -s -H $AUTH_HEADER https://api.cloudscale.ch/v1/custom-images/import | jq -r '.[].status')" != "success" ]; do
```
simu, line 189: This won't work if there's already another image uploaded in the same cloudscale.ch organization since this command lists all upload statuses and compares a potential multi-line value with the string "success".

You should probably get the image uuid of the current import in the previous step with `image_uuid=$(curl ... | jq -r '.uuid')` and use it here to only check the upload status for the current import with `https://api.cloudscale.ch/v1/custom-images/import/${image_uuid}`.

haasad: Would it make more sense to just put the image deletion step at the end of the how-to? Like that you also wouldn't have to wait for the image import to finish (takes like 2 min).

simu: Yeah, assuming we generally will spend >=2 minutes until we actually need the image, I'm fine with just cleaning up at the end of the how-to (we can probably actually completely empty and remove the bootstrap bucket at that point).

haasad: I tested the command in a project where there were already custom images imported at an earlier stage, so it looks like they don't keep those import status  info indefinitely. IIRC I read that failed imports disappear after one week, so it's probably the same for successful imports.

corvus-ch (approved):

PR #193, opened 2022-08-05
Use cluster-admin to drain nodes, add wait timeout

haasad:
- `oc as=cluster-admin` for node drains, otherwise the commands will just happily ignore the drain and run terraform apply without draining
- add a 5min timeout for the `oc wait` for nodes being ready. It's not strictly necessary because some poddisruptionbudgets will prevent the next node drain anyway, but it reduces noise in the output


rxbn (approved):
LGTM

54nd20 (approved):
LGTM

PR #196, opened 2022-08-23
Add basis for SLO documentation and the CSI operations SLO

glrf:
This PR adds a basis to document APPUiO Managed OpenShift SLOs. Wasn't entirely sure how to write this but this should give us a basis to iterate on.

### DIFF
--- a/docs/modules/ROOT/pages/explanations/slos.adoc
+++ b/docs/modules/ROOT/pages/explanations/slos.adoc
@@ -1,0 +1,40 @@
+= Service Level Objectives
+
+APPUiO Managed OpenShift 4 comes with a collection of https://sre.google/sre-book/service-level-objectives/[service level objectives (SLOs)].
+This document defines and explains these SLOs.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/explanations/slos.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+== Storage
+
+Persistent storage is a key component of a feature complete Kubernetes cluster.
+Any storage issues directly impacts the service level for users.
+
+=== CSI Operations
+
+****
+*99.5% of all CSI operations complete successfully*
```
glrf, line 20: I pulled this number out of my arse. I defined it here https://github.com/appuio/component-openshift4-slos/blob/master/class/defaults.yml#L30. 

I have no Idea what would be a good objective. I propose we revisit all of these SLOs once he actually have them

simu (approved):

PR #206, opened 2022-09-26
Omit setting OpenShift update channel by default

simu:
This needs component openshift4-version v1.1.0 or newer.


corvus-ch (changes requested):
Don't we have a chicken and egg issue here with the dynamic facts not yet being reported?

corvus-ch (approved):

corvus-ch:
> it's handled by providing dynamic facts overrides for the initial compiles,

Ah, this is what I was missing.

PR #220, opened 2022-12-12
Make Cilium the default

corvus-ch:
Dropped any references of Cilium being optional.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/cloudscale/install.adoc
@@ -12,11 +12,6 @@ These steps follow the https://docs.openshift.com/container-platform/latest/inst
 https://www.terraform.io[Terraform] is used to provision the cloud infrastructure.
 --
 
-[CAUTION]
-This guide describes the installation with https://cilium.io/[Cilium] as the network provider.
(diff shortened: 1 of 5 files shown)


megian:
There are references to OpenShift SDN and OVNKubetetes in the description. Should this be included in this PR?

simu (changes requested):
Definitely needs more changes, at least switching the installer config to `networkType: Cilium` instead of patching it in the next step

corvus-ch:
> There are references to OpenShift SDN and OVNKubetetes in the description. Should this be included in this PR?

* In the context of vSphere, OVN is still mentioned. Its about not using a certain network range when OVN is used.
* On the reference about operators, openshift-sdn is mentioned as a namespace managed by the network operators.

They both are valuable information I would not like to drop right now. I see them relevant for existing clusters still using them.

megian:
> > There are references to OpenShift SDN and OVNKubetetes in the description. Should this be included in this PR?
> 
>     * In the context of vSphere, OVN is still mentioned. Its about not using a certain network range when OVN is used.
> 
>     * On the reference about operators, openshift-sdn is mentioned as a namespace managed by the network operators.
> 
> 
> They both are valuable information I would not like to drop right now. I see them relevant for existing clusters still using them.

Fine by me.

simu (approved):

PR #230, opened 2023-02-24
Add pages documenting common labels and annotations on OpenShift 4

simu:


corvus-ch (approved):
Nice one, thanks.

What about the run-book URL? Is that also a label, or is that an annotation? Should we have a documentation of common annotations, or does this already exist?

simu:
> What about the run-book URL? Is that also a label, or is that an annotation? Should we have a documentation of common annotations, or does this already exist?

That's an annotation (I briefly toyed with the idea to make this page into "Common labels and annotations" but decided against it). I'll add a separate page, since it's not documented anywhere either.

bastjan (approved):
Nice 🚀

anothertobi (approved):

PR #231, opened 2023-02-28
Decide how to trigger unattended upgrades for OCP4

simu:
This isn't final, and may get updated after internal discusssions

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/maintenance-trigger.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/maintenance-trigger.adoc
@@ -99,8 +99,19 @@ However, by implementing our own controller, we should be able to add support fo
 
 == Decision
 
-TBD
+We decided to implement our own upgrade controller.
 
 == Rationale
 
-TBD
+By implementing our own upgrade controller we strike a good balance between implementation effort, customizability, complexity of monitoring the upgrade process.
+
+With a controller that runs on the cluster which is being upgraded, we build a triggering mechanism which can easily react to dynamic changes on the cluster, such as skipping a scheduled upgrade if the cluster is unhealthy.
+In comparison, a solution where (parts of) the triggering mechanism run outside the target cluster which is being upgraded, would need a complex feedback loop across multiple systems to be able to react to dynamic changes on the cluster.
+
+Additionally, with a controller that runs in the cluster, we don't need to invest time into engineering new monitoring infrastructure for external systems, such as GitLab CI pipelines which run Renovate or the Commodore catalog compilation CI pipeline, in order to be able to monitor the upgrade triggering mechanism itself.
+
+Further, by writing our own controller, we're able to accommodate specific upgrade requirements such as announcing target versions ahead of time or custom upgrade schedules more easily than by going with any of the other options.
+
+On the implementation side -- since we've got significant experience in writing Kubernetes controllers in Go -- the overhead of bootstrapping and maintaining our own dedicated tool to trigger upgrades is acceptable.
+
+Finally, having the update triggering mechanism running in the cluster matches RedHat's upgrade concept for OpenShift 4 which is inherently cluster-scoped, see the https://docs.openshift.com/container-platform/latest/updating/index.html[upstream documentation].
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/explanations/decisions/maintenance-trigger.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+== Rationale
+
+Renovate's approach of creating and automerging MRs to update the desired OpenShift 4 version for individual clusters fits well with our current GitOps approach.
+
+By engineering the upgrade triggering mechanism to interact with the cluster's Project Syn config, we can keep the current machinery based on Project Syn to set the cluster's desired version in the `ClusterVersion` resource unchanged.
+Additionally, we already understand how to engineer extensions for Renovate to create MRs for dependencies which aren't supported out of the box.
+
+Finally, we already have experience with scheduling automerging Renovate MRs for a certain time.
+Since we control Renovate default configuration through self-hosting our custom Renovate, we can set a global default trigger time for upgrades, while still allowing custom start times for tenants or clusters through `renovate.json` in individual tenant repos.
```
simu, line 59: Not final yet, but personally, I'm leaning towards Renovate at the moment.

DebakelOrakel (approved):
Looks good to me

anothertobi:
Although I very much like the idea of triggering updates through Renovate, I fear that this increases complexity a lot by relying on a very specific setup (automated catalog compile and push on commit in a tenant repo).

I'd prefer if an active component in the cluster would update the ClusterVersion object. With this approach we would be more flexible should there be any changes in the future to the update system of OpenShift (per cluster auth against Red Hat API or similar). Monitoring the updates of the ClusterVersion object through an in-cluster component would be fairly easy compared to the alternative of implementing monitoring for failed GitLab jobs.

DebakelOrakel:
> Although I very much like the idea of triggering updates through Renovate, I fear that this increases complexity a lot by relying on a very specific setup (automated catalog compile and push on commit in a tenant repo).
> 
> I'd prefer if an active component in the cluster would update the ClusterVersion object. With this approach we would be more flexible should there be any changes in the future to the update system of OpenShift (per cluster auth against Red Hat API or similar). Monitoring the updates of the ClusterVersion object through an in-cluster component would be fairly easy compared to the alternative of implementing monitoring for failed GitLab jobs.

That means `ocp-drain-monitor` is not `ocp-maint-monitor` but `ocp-maint-operator` :D

simu:
> Although I very much like the idea of triggering updates through Renovate, I fear that this increases complexity a lot by relying on a very specific setup (automated catalog compile and push on commit in a tenant repo).
> 

That's a good point, and as you mention below, monitoring this chain of events to alert an engineer if an update wasn't triggered correctly would be quite complex to implement.

> I'd prefer if an active component in the cluster would update the ClusterVersion object. With this approach we would be more flexible should there be any changes in the future to the update system of OpenShift (per cluster auth against Red Hat API or similar). Monitoring the updates of the ClusterVersion object through an in-cluster component would be fairly easy compared to the alternative of implementing monitoring for failed GitLab jobs.

I think having an active component in the cluster would be a good approach for some aspects of triggering updates, apart from the points you mention, it would also allow us to verify the cluster health before triggering the update. If we go that route, we'd probably want to implement our own controller/operator, since I don't think it would be easy to get the changes we need upstreamed into the openshift-managed-upgrade-operator.

simu:
Updated decision page:
* Added new goals (pre-maintenance health check, maintenance trigger monitoring)
* Added new option to implement our own controller
* Removed decision and rationale for now

simu:
For the review: please check if I forgot to include a point we discussed in the "Rationale" section

anothertobi (approved):

anothertobi:
The good 10s after approving thoughts: You might want to add that a component running in the cluster fits in well with Red Hat's upgrade concept, which is cluster based.

simu:
> The good 10s after approving thoughts: You might want to add that a component running in the cluster fits in well with Red Hat's upgrade concept, which is cluster based.

Added

DebakelOrakel (approved):

corvus-ch (approved):

PR #239, opened 2023-05-02
Add decision page for sharing ArgoCD between teams

simu:


docs/modules/ROOT/pages/explanations/decisions/syn-argocd-sharing.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+With this approach, the team responsible for the cluster itself would continue using the current ArgoCD project (`syn`) and root application (`root`).
+This team can bootstrap ArgoCD projects and root applications for any other teams who deploy applications to the cluster through Project Syn.
+
+To support this approach, the argocd component library needs to be made "team-aware" (similar to the first alternative).
+Instead of using the component instance to team mapping to add labels to each ArgoCD application, the component library can ensure the correct ArgoCD project is configured for each application.
```
glrf, line 40: I'm not entirely sure how this would fix the pausing auto sync issue. Wouldn't we still need to

pause root app -> pause team app -> pause component app

If I understand the approach correctly the root app will still deploy the "team" app so to disable auto sync for the team app we'd still need to disable auto sync for the root. Or should the root app not manage sync options of the team app?

simu: The idea is that each team app would be deployed as an additional root app in a separate project (i.e. the team app for `team1` would point to a separate directory in the cluster catalog where the team app itself is also stored) -- maybe that's not quite clear from the current text. Also, I'm not 100% sure how the bootstrap procedure for this will look like.

glrf: Aahh. So we'd bootstrap a completely different root app that just happens to live in the same cluster catalog.

I definitely see the appeal of separating this now. Seems like a reasonable approach. We just need to make sure that we get the boostrap procedure right. There IMO shouldn't be any manual steps but it should be possible to boostrap such a "team" project through the config hierarchy, even if we wouldn't completely manage it after creation.

simu:
> But IMO should probably actually be an SDD

I agree, I'll write an extended version as a full design doc (after we reach an agreement on how to proceed) since there's some tricky bits in the actual implementation.

lieneluksika (approved):

corvus-ch (commented):
Does this include the ability to deploy changes without the risk of also applying changes to things in the responsibility of another team? If I understand correctly, we still have one single catalog compiled as a whole. Are we OK with this, or would it be desirable, to have compilation happen on a team basis?

Idea:
* Keep one repository
* Compile catalog according to team responsibilities
* Commit changes to per team branches

simu:
> Does this include the ability to deploy changes without the risk of also applying changes to things in the responsibility of another team? If I understand correctly, we still have one single catalog compiled as a whole. Are we OK with this, or would it be desirable, to have compilation happen on a team basis?

As discussed: supporting partial catalog compilation is a larger task (needs some changes in Commodore), so we're making this a non-goal for this decision. However, as noted in the text, the chosen approach should work mostly unchanged if we decide to switch to a different model for catalog compilation which allows partial catalog compilation.

corvus-ch (approved):

PR #242, opened 2023-05-04
Add decision page for multiple argocd instances

DebakelOrakel:


docs/modules/ROOT/pages/explanations/decisions/multi-instance-argocd.adoc
```diff
@@ -0,0 +1,50 @@
+= Multiple ArgoCD Instances
+
+== Problem
+
+If a customer also want's to use ArgoCD for their deployment, he currently has to install ArgoCD with the same version we use and not install any CRDs.
```
simu, line 5:
```suggestion
If a customer also wants to use ArgoCD for their deployment, they currently have to install ArgoCD with the same version we use and not install any CRDs.
```

docs/modules/ROOT/pages/explanations/decisions/multi-instance-argocd.adoc
```diff
@@ -0,0 +1,50 @@
...
+* Separation of concerns
+* Separation of responsibility / accountability
+
+=== Make ArgoCD component multi-instance ready
+
+Making the ArgoCD component multi-instance ready would give us the option to deploy multiple instances of ArgoCD and controlling the version and configuration.
+
+=== Use ArgoCD operator and update ArgoCD component to use operator
+
+Using the ArgoCD operator would release us from managing customers instances. The customer would gain the ability to deploy as many instances as he'd like without any conflict with existing deployments.
```
simu, line 28:
```suggestion
Using the ArgoCD operator enables us to offer customers a mechanism with which they can create as many ArgoCD instances as they need.
This is an improvement over the current state and the other proposals which would still require us to create and manage the customer's instances.
```

docs/modules/ROOT/pages/explanations/decisions/multi-instance-argocd.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+=== Create separate ArgoCD project and root application for customer
+
+Using the same ArgoCD instance for managing the cluster and the customers isn't an option for various reasons:
+* Separation of concerns
+* Separation of responsibility / accountability
+
+=== Make ArgoCD component multi-instance ready
+
+Making the ArgoCD component multi-instance ready would give us the option to deploy multiple instances of ArgoCD and controlling the version and configuration.
```
simu, line 24:
```suggestion
Making the ArgoCD component multi-instance ready enables us to deploy multiple instances of ArgoCD while ensuring that there's no version mismatch between the CRD and the installations.
Additionally, we could manage the ArgoCD configuration for all of the instances if desired.
```

docs/modules/ROOT/pages/explanations/decisions/multi-instance-argocd.adoc
```diff
@@ -0,0 +1,50 @@
...
+== Problem
+
+If a customer also want's to use ArgoCD for their deployment, he currently has to install ArgoCD with the same version we use and not install any CRDs.
+This process needs synchronization between us and the customer, if there is a new version of ArgoCD to be rolled out.
+
+=== Goals
+
+* We can deploy multiple instances of ArgoCD
+* No synchronization needed between us and customer, if an update needs to be rolled out
+
```
simu, line 12:
```suggestion

=== Non-Goals

* Allow multiple different versions of ArgoCD to be installed on the same cluster
* Provide VSHN-managed ArgoCD instances to the customer
```

Is the second bullet actually a non-goal? In my opinion it is, since this should explicitly be a goal if we want to optionally offer that we manage the customer's instance config for them.

docs/modules/ROOT/pages/explanations/decisions/multi-instance-argocd.adoc
```diff
@@ -0,0 +1,50 @@
...
+This option would give us the highest degree of versatility in the future.
+
+
+== Decision
+
+Use ArgoCD operator.
+
+Using the operator gives us the most versatility.
+Implementation can be done in multiple steps, steward and component-argocd can be adopted at a later point.
```
simu, line 40:
```suggestion
Implementation can be done in multiple steps, Steward and the ArgoCD Commodore componenT can be migrated to use the operator at a later point.
```

docs/modules/ROOT/pages/explanations/decisions/multi-instance-argocd.adoc
```diff
@@ -0,0 +1,50 @@
...
+When creating component-argocd-operator we need to make sure to remove 2 of the CRDs that come with the operator.
+
+Ideally the operator should be handled by steward / component-argocd in the future.
```
simu, line 43:
```suggestion
Ideally the operator should be installed by Steward and the ArgoCD Commodore component in the future.
```

corvus-ch (approved):
With the changes by @simu applied, this is looking goot to me.

PR #243, opened 2023-05-09
Initial vSphere IPI install instructions

simu:
This should not be considered final, but I think it's now good enough to review and merge. We can then do another pass over it once we setup the next cluster.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
@@ -182,6 +182,4 @@ include::partial$install/registry-samples-operator.adoc[]
 
 include::partial$install/finalize_part1.adoc[]
 
-//include::partial$install/registry-acl-fix.adoc[]
-
 include::partial$install/finalize_part2.adoc[]
--- a/docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
+++ b/docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
@@ -1,0 +1,189 @@
+= Install OpenShift 4 on vSphere
+:ocp-minor-version: 4.11
+:k8s-minor-version: 1.24
+:ocp-patch-version: {ocp-minor-version}.9
+:provider: vsphere
+
+[abstract]
+--
+Steps to install an OpenShift 4 cluster on VMWare vSphere.
+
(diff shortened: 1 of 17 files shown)


docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
```diff
@@ -0,0 +1,189 @@
...
+export VSPHERE_PASSWORD=<password>
+----
+
+.vSphere resources
+[source,bash]
+----
+export VSPHERE_CLUSTER=<cluster name>
+export VSPHERE_DATACENTER=<datacenter name>
+export VSPHERE_DATASTORE=<datastore name>
+export VSPHERE_FOLDER=<folder name>
```
haasad, line 67: Is this required? Don't see it referenced anywhere else?

docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
```diff
@@ -0,0 +1,189 @@
...
+<2> This command expects that your SSH config is setup so that `ssh ${JUMPHOST_FQDN}` works without further configuration
+<3> The `openshift-install` tool respects the `https_proxy` environment variable
+====
+
+. Trust the vSphere CA certificate
++
+.Ubuntu 22.04
+[source,bash]
+----
+curl -kLo vsphere-ca.zip ${VCENTER_API_URL}/certs/download.zip <1>
```
haasad, line 156: Footnote is missing here afaics

haasad (approved):
LGTM in the current form :+1: we'll have to refine it once we set up the next cluster following these instructions, but looks already very complete

docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
simu: It's not required by default. I've removed it for now.

docs/modules/ROOT/pages/how-tos/vsphere/install.adoc
simu: I figured out how to tell curl to resolve DNS names over a SOCKS5 proxy, so I've updated this section slightly.

corvus-ch (approved):
While it is not necessarily easy to maintain, I really like how the includes helps to understand what the common steps are and where the different platforms differ.

PR #247, opened 2023-05-24
Add `syn_team` to common labels

DebakelOrakel:
This label is used to route alerts to the responsible team.


corvus-ch (approved):

simu (approved):
Mentioning the concept of a component doesn't make sense when defining a label for individual alerts. See inline comment for my suggested wording.

PR #255, opened 2023-06-16
Managed Upgrade Controller: Introduce Hooks

bastjan:
Introduces the concept of `UpgradeJobHook`s and removes the `webhook` idea.


simu (approved):
Design seems reasonable overall.

Some wording suggestions / clarifications inline

PR #257, opened 2023-06-27
Add step to enable AppCat during installation on Cloudscale

glrf:
AppCat should now be enabled by default on APPUiO Managed OpenShift on Cloudscale. This simply requires adding a single class and will reuse the Cloudscale token stored under `${cluster:tenant}/${cluster:name}/cloudscale/token` to provision buckets on cloudscale

### DIFF
--- a/docs/modules/ROOT/partials/install/prepare-commodore.adoc
+++ b/docs/modules/ROOT/partials/install/prepare-commodore.adoc
@@ -40,6 +40,24 @@ git push
 popd
 ----
 
+
+ifeval::["{provider}" == "cloudscale"]
+. Add AppCat to cluster configuration
++
+[source,bash]
+----
+pushd "inventory/classes/${TENANT_ID}/"
+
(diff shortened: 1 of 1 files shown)


bastjan:
Are there no plans enabling it in global defaults?

glrf:
> Are there no plans enabling it in global defaults?

I'm a bit torn if we should do that, as AFAIK you can't easily "uninclude" a class and I'd generally like to have the option to easily disable appcat on a cluster

bastjan:
> > Are there no plans enabling it in global defaults?
> 
> I'm a bit torn if we should do that, as AFAIK you can't easily "uninclude" a class and I'd generally like to have the option to easily disable appcat on a cluster

I'll ask in the daily tmrw what the preferred option for Aldebaran is

bastjan:
We decided to have it in global defaults. We do not want more installation steps and don't see a case currently where we need to remove appcat.

glrf:
Alright, in that case I'll add the class to the global defaults for OpenShift on cloudscale. If this ever gets an issue that should be easy enough to revert

PR #266, opened 2023-07-19
The minio client changed `mc policy` to `mc anonymous`

bastjan:
Functionality is the same. Tested while setting up a PoC cluster.


simu (changes requested):
Please add a required minimum version of `mc` to a version which has the new command in the prerequisites section

simu (approved):

PR #268, opened 2023-08-15
Decision page for scheduled GitLab merges

HappyTetrahedron:
Please check for NIH - I currently think that it's sensible to write our own bespoke service, given there is no existing solution that I could find and the overall complexity really shouldn't be that high.

### DIFF
--- a/docs/modules/ROOT/pages/explanations/decisions/scheduled-mr-merges.adoc
+++ b/docs/modules/ROOT/pages/explanations/decisions/scheduled-mr-merges.adoc
@@ -1,0 +1,65 @@
+= Scheduling the merge of a GitLab merge request
+
+== Problem
+
+We want to be able to merge GitLab merge requests at preset times.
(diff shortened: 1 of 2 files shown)


bastjan (approved):

docs/modules/ROOT/pages/explanations/decisions/scheduled-mr-merges.adoc
```diff
@@ -0,0 +1,63 @@
+= Scheduling the merge of a GitLab merge request
+
+== Problem
+
+We want to be able to merge GitLab merge requests at preset times.
+This enables us to roll out changes during maintenance windows in an automated fashion.
```
simu, line 6: I'd add a short paragraph here noting that while there's a GitLab feature request for scheduled merges, there's currently no suitable functionality out of the box.

simu (approved):
LGTM overall

PR #277, opened 2023-09-21
Add decision page for managing the admin kubeconfig

simu:


docs/modules/ROOT/pages/explanations/decisions/admin-kubeconfig.adoc
```diff
@@ -0,0 +1,99 @@
...
+* Define a method to manage emergency access credentials for OpenShift 4 clusters
+* The credentials should be relatively short-lived and it must be possible to rotate them
+
+=== Non-Goals
+
+* Replace regular authentication
+
+== Proposals
+
+=== Credential type
```
chloesoe, line 23: We have two approaches for "Credentail type" and three approaches for "Credential managment" correct?

chloesoe:
```suggestion
=== Credential type

For the credential type there are two approaches.
```

Short introduction below the section heading helps the reader to be prepared to  the following paragraphs.

docs/modules/ROOT/pages/explanations/decisions/admin-kubeconfig.adoc
```diff
@@ -0,0 +1,99 @@
...
+As the name suggests, these tokens don't expire.
+
+The only way to permanently invalidate service account tokens (both non-expiring and time-bound) is to delete the service account.
+Creating a new service account with the same name in the same namespace doesn't reactivate tokens associated with a previous service account, since the tokens contain the service account's Kubernetes resource UID.
+
+The proposed approach for using service account tokens is to use the TokenRequest API to create short-lived API tokens to generate expiring admin credentials by default.
+Additionally, introduce a mechanism to force the tool to recreate the service account to invalidate any old tokens that might have leaked.
+That mechanism might be as simple as having the tool reconcile the service account and recreate it if it gets deleted.
+
+=== Credential management
```
chloesoe, line 55:
```suggestion
=== Credential management

For credential management there are three options
```

Same as above, perhaps we should number the options here and at credential type ?

docs/modules/ROOT/pages/explanations/decisions/admin-kubeconfig.adoc
```diff
@@ -0,0 +1,99 @@
...
+This would provide some level of separation of concerns, since managing admin credentials isn't necessarily part of the Project Syn bootstrap process.
+Additionally, having a separate tool allows us to have releases independent of the fairly complex Steward release process.
+
+==== Manage credentials by hand
+
+Another approach is to manage and renew the admin credentials by hand.
+
+== Decision
+
+Use service account tokens generated through the TokenRequest API and implement a custom controller to manage the service account, cluster role binding and tokens.
```
chloesoe, line 78: So, I guess that the decision consist of one approach from "credential type" and one approach from "credential management", correct? 

We should make it clearer here, which approach for credential management is taken. I did not find out which one directly

docs/modules/ROOT/pages/explanations/decisions/admin-kubeconfig.adoc
simu: I've updated the text to have an overall introduction in section `Proposals`, with short sentences reiterating the intro for each sub-section

docs/modules/ROOT/pages/explanations/decisions/admin-kubeconfig.adoc
simu: "implement a custom controller to manage the service account, cluster role binding and tokens" seems pretty clear to me

chloesoe: yeah, my approach no deep expertise while reading is to ctrl+F and look for "custom controller" and would not find a match above, because there it's named "new controller".  So it would make it clearer, if above instead of "new controller" we would name it "new custom controller".  I'll add a proposal with links.

chloesoe: https://github.com/appuio/openshift4-docs/pull/277/commits/2178830abc8dee1f493228cc26029adaa5d92a1f

chloesoe:
lgtm, thanks for your work here

docs/modules/ROOT/pages/explanations/decisions/admin-kubeconfig.adoc
simu: seems overkill for slightly more than a screen page of text, and the links don't even work

chloesoe: true, they somehow do not work... now you changed the title, it should be clear...

chloesoe: ah, that's how it works. not that obvious to find in https://docs.asciidoctor.org/asciidoc/latest/macros/xref/ within seconds :see_no_evil: 
thanks

corvus-ch (approved):

chloesoe (approved):

bastjan (approved):

PR #297, opened 2023-12-15
Add Exoscale architecture reference documentation

DebakelOrakel:


docs/modules/ROOT/assets/images/ocp4-architecture-exoscale.svg
corvus-ch, line 1: It should be clear that other systems can be both inside and outside the Exoscale zone. Also, make it clear that the egress traffic does NOT go through an internet gateway.

The Rook Ceph storage cluster is missing.

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
```diff
@@ -0,0 +1,107 @@
...
+
+
+== Networking
+
+=== Security Groups
+
+On {infra-type} APPUiO Managed OpenShift 4 uses public ips for each node in the cluster.
+See https://kb.vshn.ch/oc4/explanations/exoscale/limitations.html#_private_networks[Limitations] of the {infra-type} environment.
+
+The individual VMs are placed in https://community.exoscale.com/documentation/compute/security-groups[Security Groups] to restrict access and isolate the nodes from the public internet.
```
corvus-ch, line 27: Elaborate on the user facing consequence: source IP of outgoing request is that of the individual node. Nodes can come and go. Thus, those IPs are no reliable source for firewall rules and other security policies.

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
```diff
@@ -0,0 +1,107 @@
...
+2. The "Ingress VIP" for the OpenShift Ingress Router
+APPUiO Managed OpenShift 4 uses a public floating IP as the Ingress VIP.
+
+APPUiO Managed OpenShift 4 uses two Load Balancer instances to manage the API and ingress VIPs and distributes traffic to the master / infrastructure nodes.
+
+=== Pod and service networks
+
+include::partial$architecture/networking-pods.adoc[]
+
+=== Exposing the cluster
```
corvus-ch, line 44: This is a somewhat strange structure. The subject of ingress has now cum up the third time in this document (requirements, VIPs and now this). Further, this is written like requirements (usage of words like must) instead of just reflecting the state as is.

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
```diff
@@ -0,0 +1,107 @@
...
+The PROXY protocol should be enabled to preserve source IPs.
+Forwarding should happen transparently in TCP mode.
+VSHN will manage a wildcard DNS record pointing to this IP.
+Additional DNS records can be pointed to this IP by the customer.
+
+=== External services
+
+include::partial$architecture/networking-external.adoc[]
+
+== Storage
```
corvus-ch, line 65: The Rook Ceph Storage cluster is missing.

docs/modules/ROOT/partials/nav.adoc
```diff
@@ -14,6 +14,7 @@
 ** xref:oc4:ROOT:references/architecture/single_sign_on.adoc[]
 
 ** Exoscale
+*** xref:oc4:ROOT:references/exoscale/architecture.adoc[Architecture]
```
corvus-ch, line 17:
```suggestion
** xref:oc4:ROOT:references/exoscale/architecture.adoc[Exoscale]
```

docs/modules/ROOT/partials/nav.adoc
```diff
@@ -39,6 +40,7 @@
 *** xref:oc4:ROOT:how-tos/cloudscale/decommission.adoc[Decommissioning]
 
 ** Exoscale
+*** xref:oc4:ROOT:references/exoscale/architecture.adoc[Architecture]
```
corvus-ch, line 43:
```suggestion
```

Or do the same for vSphere for consistency. I prefer to not have the same page at multiple locations in the navigation. It is confusing.

corvus-ch (changes requested):
Please fix the following aspects:

* Make it more obvious that on Exoscale nodes are directly connected to the internet and that outgoing (egress) connections have a nodes IP as its source which can change over time.
* Add the Rook Ceph storage cluster which is mandatory for Exoscale clusters.

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
```diff
@@ -0,0 +1,107 @@
...
+2. Ingress load balancer: Provides an ingress point for application traffic flowing in from outside the cluster.
+
+See the https://docs.openshift.com/container-platform/latest/installing/installing_bare_metal/installing-bare-metal.html#installation-requirements-user-infra_installing-bare-metal[upstream documentation] for details on {infra-type} requirements.
+
+
+== Networking
+
+=== Security Groups
+
+On {infra-type} APPUiO Managed OpenShift 4 uses public ips for each node in the cluster.
```
simu, line 24:
```suggestion
On {infra-type}, APPUiO Managed OpenShift 4 uses public IPs for each node in the cluster.
```

docs/modules/ROOT/partials/nav.adoc
simu: The current state is that we have links to many (most?) pages in kb.vshn.ch/oc4 in multiple places in the navigation, since we restructured the navigation. Also, given the current navigation structure, there's no real way to avoid this (as far as I can tell/remember from the last time I looked at it)

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
simu: Looks like that's covered in the partial.

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
```diff
@@ -0,0 +1,107 @@
...
+
+== Architecture overview
+
+include::partial$architecture/overview.adoc[]
+
+== {infra-type} requirements
+
+APPUiO Managed OpenShift 4 on {infra-type} needs a https://docs.openshift.com/container-platform/4.14/installing/installing_bare_metal/installing-bare-metal.html#installation-load-balancing-user-infra_installing-bare-metal[Load Balancer setup] that must meet the following requirements:
+
+1. API load balancer: Provides a common endpoint to interact with the OpenShift and Kubernetes.
```
simu, line 13: either

```suggestion
1. API load balancer: Provides a common endpoint to interact with the OpenShift and Kubernetes APIs.
```

or

```suggestion
1. API load balancer: Provides a common endpoint to interact with OpenShift and Kubernetes.
```

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
```diff
@@ -0,0 +1,107 @@
...
+
+2. Ingress load balancer: Provides an ingress point for application traffic flowing in from outside the cluster.
```
simu, line 15: I'd probably use "endpoint" here for consistency with the description of the API LB

```suggestion
2. Ingress load balancer: Provides an endpoint for application traffic flowing in from outside the cluster.
```

docs/modules/ROOT/pages/references/exoscale/architecture.adoc
simu: We can probably leave out this section completely for Exoscale. This is important for vSphere (or other on-prem setups) where the customer needs to expose the cluster in their environment, but for Exoscale/cloudscale.ch/... this is an implementation detail that's (IMO) sufficiently covered in the previous sections.

The one bit of information that's relevant for Exoscale is that we provide a CNAME target record for the customer to point additional DNS records to.

docs/modules/ROOT/partials/architecture/storage.adoc
```diff
@@ -19,16 +19,27 @@ They're allocated dynamically based on requests from workloads (applications or
 These block devices are automatically attached to the VM hosting the application container.
 They're deleted when the corresponding Kubernetes `PersistentVolume` resource is deleted.
 
+ifeval::["{infra-type}" != "Exoscale"]
 The {infra-type} CSI driver is the in-cluster component which is responsible for allocating, attaching and deleting the persistent volume block devices.
+endif::[]
+
+ifeval::["{infra-type}" == "Exoscale"]
+NOTE: On the {infra-type} environment there is no Kubernetes integration available for such block devices.
+https://products.vshn.ch/appuio/managed/storage_cluster.html[APPUiO Managed Storage Cluster] is available as and addon to the cluster offering if such functionality is required.
```
corvus-ch, line 28:
```suggestion
IMPORTANT: {infra-type} does not provide storage usable by Kubernetes as persistent volumes.
To fill this gap, {product} in {intra-type} uses https://products.vshn.ch/appuio/managed/storage_cluster.html[APPUiO Managed Storage Cluster] to provide storage to be used as read write once and read write many persistent volumes.
```

docs/modules/ROOT/partials/nav.adoc
corvus-ch, line 17:
```suggestion 
** xref:oc4:ROOT:references/exoscale/architecture.adoc[Exoscale]
```

We aim to get rid of those empty menu entries.

corvus-ch (changes requested):
The VSHN managed storage cluster is a requirement and not an optional add-on.
Plus, we should get rid of those empty menu entries where ever possible. At least we should not introduce new ones.

corvus-ch (approved):

PR #301, opened 2024-01-16
Remove LDAP fro mexternal dependencies

corvus-ch:


simu:
Change is already incorporated on master branch

PR #313, opened 2024-03-07
Add admin certificate invalidation to install

bastjan:
Also adds check for emergency credentials.


docs/modules/ROOT/partials/install/finalize_part1.adoc
```diff
@@ -72,27 +72,40 @@ done
...
 ====
-You may see the error `Unable to connect to the server: x509: certificate signed by unknown authority` when executing `kubectl` or `oc` commands after the cluster's production API certificate has been deployed by Project Syn.
+You need to be in the passbolt group `VSHN On-Call`.
+
+If the command fails, check if the controller is already deployed, running, and if the credentials are uploaded:
 
-This error can be addressed by removing the initial CA certificate data from the admin kubeconfig as shown in this step.
+[source,bash]
+----
+kubectl --as=system:admin -n appuio-emergency-credentials-controller get emergencyaccounts.cluster.appuio.io -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.status.lastTokenCreationTimestamp}{"\n"}{end}'
```
haasad, line 93: is there as specific reason to use `--as=system:admin` here?

haasad (approved):
would it make sense to include a step that encourages you to first also check your personal (non-admin) login to ensure the standard login method also works?

bastjan:
> would it make sense to include a step that encourages you to first also check your personal (non-admin) login to ensure the standard login method also works?

I think it is valid to setup the "standard" login independently. Might be due to special needs or whatever. Until then it is still better to use emergency-credentials than passing around the admin kubeconfig.

docs/modules/ROOT/partials/install/finalize_part1.adoc
bastjan: nope copy-paste oversight

haasad:
yes, that makes sense

PR #314, opened 2024-03-12
Architecture proposal for pausing MCP upgrades

bastjan:


docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
```diff
@@ -136,15 +136,41 @@ If any of the operators is degraded, the cluster should be considered unhealthy
 
 It should be possible to exclude operators.
 
+=== The controller can pause machine configuration pools to delay node reboots [[mcp-upgrade-pause]]
+
+This allows to update master nodes and operators during office hours without affecting workload on the worker nodes.
+
+The UpgradeJob is marked as paused if all conditions of <<upgrade-success>> are met but there are paused machine configuration pools.
+
+The overall upgrade timeout (`.spec.timeout`) is unaffected by the pause and continues to count down.
```
haasad, line 145: This means we would need to extend the upgradeTimeout compared to what we're using now, correct? 12h wouldn't be enough if we start at 10:00 and want a `delayUpgrade.startAfter` of 12h for unpausing at 22:00. Just making sure I understand it correctly.

docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
```diff
@@ -136,15 +136,41 @@ If any of the operators is degraded, the cluster should be considered unhealthy
 
 It should be possible to exclude operators.
 
+=== The controller can pause machine configuration pools to delay node reboots [[mcp-upgrade-pause]]
```
haasad, line 139:
```suggestion
=== The controller can pause and unpause machine configuration pools to delay node reboots [[mcp-upgrade-pause]]
```
```suggestion
=== The controller can temporarily pause machine configuration pools to delay node reboots [[mcp-upgrade-pause]]
```
With the current description it's not fully clear that the upgrade controller handles both pausing and unpausing IMO.

docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
```diff
@@ -221,6 +247,12 @@ spec:
           - openshift-monitoring
           customQueries:
           - query: "up{job=~"^argocd-.+$",namespace="syn"} != 1"
+        machineConfigPools: <8>
+        - matchLabels:
+            name: x-app-night-maintenance
+          delayUpgrade:
+            startAfter: "1h"
+            startBefore: "2h"
```
haasad, line 255:
```suggestion
            delayMin: "1h"
            delayMax: "2h"
```
`startAfter` and `startBefore` are dates in the UpgradeJob, I find it confusing if the same keys are used for relative times here

haasad (approved):

docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
bastjan: `delayMin/Max` that was also my first idea and then went forth and back 😅

docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
bastjan: Yes correct. I think it is easier to understand this way and also could not really think of a good alternative.

We might want to add a `partialUpdate` (needs better naming) at some point that only looks at CV and masters.

haasad: IIRC we explicitly said the goal of this feature is to delay it by short amounts of time only (i.e. from daytime to nighttime) and not for longer periods (multiple days), so increasing the overall `upgradeTimeout` slightly is totally fine.

PR #322, opened 2024-04-09
Use amtool in the alertmanager pod instead of silencing job

haasad:
Reduces the number of manual steps needed (no cleanup of jobs etc.) and reduces the wall of code in the how-to


simu (approved):
LGTM.

There's probably other how-tos (e.g. storage node replacement) which use the same K8s job approach for creating a silence.

haasad:
I created partials for creating and expiring silences with `amtool`. I don't have the capacity at the moment to migrate all existing silence jobs to the new method. But like this we can reuse the new method when needed.

PR #326, opened 2024-05-02
Adjust install and Cilium migration how-tos to work with pre-configured Cilium `k8sServiceHost` Helm value

simu:
The new install instructions assume that the following configuration is present somewhere in the Project Syn hierarchy:

```
parameters:
  cilium:
    cilium_helm_values:
      k8sServiceHost: api-int.${openshift:baseDomain}
      k8sServicePort: "6443"
```

See  https://git.vshn.net/syn/commodore-defaults/-/merge_requests/1789 for the VSHN config


bastjan (approved):
docs seem to be way clearer than before! 😊

HappyTetrahedron (approved):
Looks good, makes sense, didn't even find typos

PR #334, opened 2024-06-10
Document migration to KubeProxy Replacement with cilium

DebakelOrakel:
How-to which describes the steps to configure kubeProxyReplacement=strict/yes on OpenShift 4.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/network/cilium-kubeproxy-disabled.adoc
+++ b/docs/modules/ROOT/pages/how-tos/network/cilium-kubeproxy-disabled.adoc
@@ -1,0 +1,102 @@
+= Configure KubeProxy Replacement with Cilium
+
+== Prerequisites
+
+* `cluster-admin` privileges
+* Cluster is running Cilium.
+See the xref:how-tos/network/migrate-to-cilium.adoc[migrating to Cilium] how-to for a guide to migrate a cluster to Cilium.
+* `kubectl`
+* `jq`
+* Working `commodore` command
+
+== Prepare for configuration
+
+IMPORTANT: Make sure that your `$KUBECONFIG` points to the cluster you want to migrate before starting.
+
+. Select cluster
++
+[source,bash]
+----
+export CLUSTER_ID=c-cluster-id-1234 <1>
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/how-tos/network/cilium-kubeproxy-disabled.adoc
```diff
@@ -0,0 +1,102 @@
...
+----
+<1> We recommend switching to an empty directory to run this command.
+Alternatively, switch to your existing directory for the cluster.
+
+. Update `kubeProxyReplacement` for cilium
++
+[source,bash]
+----
+pushd inventory/classes/"${TENANT_ID}"
+yq -i ".parameters.cilium.cilium_helm_values.kubeProxyReplacement = true" \
```
simu, line 76: Looking at https://github.com/cilium/cilium/blob/6a1222d46652ab4d962df848605aeb6b88476311/install/kubernetes/cilium/values.yaml#L1834 this should be a string "true":

```suggestion
yq -i '.parameters.cilium.cilium_helm_values.kubeProxyReplacement = "true"' \
```

docs/modules/ROOT/pages/how-tos/network/cilium-kubeproxy-disabled.adoc
```diff
@@ -0,0 +1,102 @@
...
+. Get local cluster working directory
++
+[source,bash]
+----
+commodore catalog compile "$CLUSTER_ID" <1>
+----
+<1> We recommend switching to an empty directory to run this command.
+Alternatively, switch to your existing directory for the cluster.
+
+. Update `kubeProxyReplacement` for cilium
```
simu, line 71: Since we customize a number of features for the current `kubeProxyReplacement=partial` setup (cf. https://github.com/projectsyn/component-cilium/blob/80022c0753777a94514653822dbe8ffbf87725b6/class/defaults.yml#L34-L50), we may need to adjust slightly more here.

IIRC, we need to at least also set `parameters.cilium.cilium_helm_values.nodePort.enableHealthCheck=true` when we want to fully replace kube-proxy

simu (approved):

PR #336, opened 2024-06-19
Document `UpgradeSuspensionWindow`

bastjan:
Also see https://github.com/appuio/openshift-upgrade-controller/issues/62.

### DIFF
--- a/docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
@@ -235,7 +235,7 @@ Matching `UpgradeConfig` objects won't create `UpgradeJob` objects during the ti
 Matching `UpgradeJob` objects won't start the upgrade during the time window.
 Skipped jobs will be marked as successful with reason skipped.
 Success and finish hooks will be executed as normal.
-If the job was from a config object, the object creates a new job with the current (possibly same) version in the next normal time window.
+If the job was owned by a upgradeconfig object, the object creates a new job with the current (possibly same) version in the next non-suspended time window.
 
 === When's an upgrade job considered successful? [[upgrade-success]]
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
```diff
@@ -222,6 +222,21 @@ openshift_upgrade_controller_cluster_version_info{channel="stable-4.14",cluster_
...
+This allows us to prevent upgrades during public holidays or other special events.
+
+A `UpgradeSuspensionWindow` object blocks upgrades for a specific time window.
+The objects it's matched against can be defined through a selector.
+
+Matching `UpgradeConfig` objects won't create `UpgradeJob` objects during the time window.
+
+Matching `UpgradeJob` objects won't start the upgrade during the time window.
+Skipped jobs will be marked as successful with reason skipped.
+Success and finish hooks will be executed as normal.
```
haasad, line 237: How is it handled currently with start hooks for a noop upgradejob? I assume start and success/finish hooks are all executed, right?

Do we need to ensure that we don't have success/finish hooks that expect certain actions to be executed in a corresponding start hook? I'm thinking about a certain customer's scale-down then scale-up requirements.

docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
```diff
@@ -222,6 +222,21 @@ openshift_upgrade_controller_cluster_version_info{channel="stable-4.14",cluster_
...
+If the job was from a config object, the object creates a new job with the current (possibly same) version in the next normal time window.
```
haasad, line 238:
```suggestion
If the job was owned by a upgradeconfig object, the object creates a new job with the current (possibly same) version in the next non-suspended time window.
```

haasad (approved):

docs/modules/ROOT/pages/references/architecture/upgrade_controller.adoc
haasad: Ok, just something that we need to be aware of before we roll this out. It introduces a new corner case where only some of the hooks are executed, thereby potentially violating some expectations on the behavior that we had previously.

bastjan: Yes. I would not expect too many of those weird jobs though. Since they should™️ be idempotent anyways.

PR #348, opened 2024-08-02
Update install instructions for OpenShift 4.15 and Cilium kube-proxy-free mode

simu:
* Bump OpenShift and K8s versions in cloudscale, OpenStack and vSphere instructions
* Add steps to allow control plane bootstrap to complete with Cilium in kube-proxy-free mode


bastjan (approved):
I like the added "wait for CRD" 😊

PR #350, opened 2024-08-09
Document how to configure cluster-scoped ArgoCD instances

HappyTetrahedron:
Is this sufficient? Let me know if I should be more thorough.


simu (approved):
This looks fine, I don't think it needs more details since it's intended for us

PR #353, opened 2024-09-10
Fix and improve Exoscale install documentation

HappyTetrahedron:


docs/modules/ROOT/pages/how-tos/exoscale/install.adoc
```diff
@@ -292,7 +302,6 @@ module "cluster" {
 EOF
 terraform apply -target "module.cluster.module.lb.module.hiera"
 ----
-
```
simu, line 295: Is this blank line deletion intentional? Doesn't this break the next bullet point?

simu (approved):

docs/modules/ROOT/pages/how-tos/exoscale/install.adoc
HappyTetrahedron: oh damn, nice find! no, wasn't intentional.

PR #355, opened 2024-10-09
Folder with infraID is created automatically by the installer

haasad:
https://docs.openshift.com/container-platform/4.15/installing/installing_vsphere/installation-config-parameters-vsphere.html#installation-config-parameters-vsphere

From the docs:
 > Optional: The absolute path of an existing folder where the user creates the virtual machines, for example, /<datacenter_name>/vm/<folder_name>/<subfolder_name>. If you do not provide this value, the installation program creates a top-level folder in the datacenter virtual machine folder that is named with the infrastructure ID.


simu (commented):
I'd like having a NOTE which says that there's an optional parameter `folder` that can be used to install the cluster in a predefined folder.

haasad:
I added a note. On Monday I tried unsuccessfully to use a folder with the project syn ID instead of the default vsphere infra ID, but I'm sure it could be achieved.

PR #365, opened 2024-11-15
Update Cloudscale install documentation to enable machine-api provider

HappyTetrahedron:
I imagine at some point we'll update the openshift4-terraform defaults for cloudscale to have infra and worker counts of 0. For now, though, I assume that this is not the case and have the install instructions explicitly leave an override in.

Once we're fully migrated, we will probably also have the machine-api-provider-cloudscale component enabled and configured by default, and that step will be superfluous as well. (There's a small bootstrapping snag here: the component config tries to access the ignition CA from the openshift4-terraform config. On the very first catalog compile, that isn't there. But I'm sure we can work around that once we get that far.)


docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
```diff
@@ -122,16 +148,58 @@ kubectl taint no -l node-role.kubernetes.io/master node.cloudprovider.kubernetes
...
+for fqdn in "${LB_FQDNS[@]}"; do
+  ssh "${fqdn}" sudo puppetctl run
+done
+----
+
+. Scale up the infra and worker machinesets
++
+[source,bash]
+----
+INFRA_NODES=3 # adjust to desired number of infra nodes
```
simu, line 179: This should be 4 I think

```suggestion
INFRA_NODES=4 # adjust to desired number of infra nodes
```

docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
```diff
@@ -146,15 +201,7 @@ done && echo -e "\nCilium config updated"
 kubectl -n cilium rollout restart ds/cilium
 ----
 
-. Review and merge the LB hieradata MR (listed in Terraform output `hieradata_mr`) and run Puppet on the LBs after the deploy job has completed
```
simu, line 149: This step can't be deleted unconditionally, it's required to ensure that the Puppet LBs have a correct configuration for the ingress backends.

Additionally, we probably need a replacement step for cloudscale which somehow sets the infra node IPs as ingress backends since Terraform doesn't have those IPs anymore.

docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
```diff
@@ -98,6 +98,32 @@ terraform apply
 
 ifeval::["{provider}" == "cloudscale"]
 . Add the DNS records for etcd shown in output variable `dns_entries` from the previous step to the cluster's parent zone
+
+. Apply the manifests for the cloudscale machine-api provider
++
+[source,bash,subs="attributes+"]
+----
+cat ../machine-api-provider-cloudscale/00_secrets.yaml  | yq '(select(document_index==0) | .stringData.token) = "'$(vault kv get -format=json clusters/kv/${TENANT_ID}/${CLUSTER_ID}/cloudscale/ | jq -r .data.data.token)'" ' | kubectl apply -f -
```
simu, line 106: This could be the following:

```suggestion
kapitan refs --reveal --refs-path ../../refs ../machine-api-provider-cloudscale/00_secrets.yaml | kubectl apply -f -
```

docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
HappyTetrahedron: It's not deleted, I just shuffled it around. It should be further up now.

docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
HappyTetrahedron: Oh, TIL

docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
```diff
@@ -132,6 +163,30 @@ terraform apply
 popd
 ----
 
+. Review and merge the LB hieradata MR (listed in Terraform output `hieradata_mr`) and run Puppet on the LBs after the deploy job has completed
```
simu, line 166: This won't be sufficient for cloudscale anymore, since this also ensures that the ingress backends are set to the infra node IPs. This isn't done anymore for cloudscale since we set `infra_count=0` and we'll need a replacement step after scaling up the infra nodes which sets their IPs in the hieradata so that the ingress HAProxy config has backends.

docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
```diff
@@ -146,15 +201,27 @@ done && echo -e "\nCilium config updated"
...
+ifeval::["{provider}" == "cloudscale"]
+. Add Infra Node IPs to LB Hieradata
 +
 [source,bash]
 ----
-for fqdn in "${LB_FQDNS[@]}"; do
-  ssh "${fqdn}" sudo puppetctl run
-done
-----
+git clone git@git.vshn.net:appuio/appuio_hieradata.git
```
HappyTetrahedron, line 209: The terraform checkout of this repo uses the project bot user via https,  and trying to do anything in it asks for the bot's password, which is annoying to script for.
I may be missing something. But I found it easier to just re-clone with ssh...

simu: You could maybe do something like

```
git remote add origin-git git@git.vshn.net:appuio/appuio-hieradata.git
```

in the Terraform checkout and then

```
git push origin-git master
```

but that's not easier than just cloning the repo again.

simu (approved):

docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
HappyTetrahedron: yeah... plus, in my local directory the terraform-managed checkout wasn't in a clean state. Whether that was a result of me doing things wrong or a normal possibility, I'm not sure. But a fresh clone certainly never has issues with unclean git trees...

PR #368, opened 2024-12-01
Architecture: Generic, template-able resource reconciler  (Espejote)

DebakelOrakel:


docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,411 @@
...
+
+"Espejote" for "big mirror" in Spanish.
+It's a play on the earlier project "Espejo" which means "mirror."
+"Espejo" was a tool to mirror resources to multiple namespaces.
+"Espejote" is a more general tool to manage arbitrary configurations in a cluster.
+
+== Implementation
+
+The controller is based on `controller-runtime`.
+It's managed through a Custom Resource Definition (CRD) called `(Cluster)ManifestTemplate`.
```
DebakelOrakel, line 39:
```suggestion
It's managed through a Custom Resource Definition (CRD) called `(Cluster)ManagedResource`.
```

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,411 @@
...
+          // namespace: c.metadata.namespace, <9>
+          labels+: {
+            'espejote.io/created-by': 'copy-configmap'
+          }
+        }
+      },
+      for c in cm.items <10>
+    ]
+----
+<1> `ManifestTemplate` are always namespace-scoped and can't access any resources outside of their namespace.
```
DebakelOrakel, line 259:
```suggestion
<1> `ManagedResource` are always namespace-scoped and can't access any resources outside of their namespace.
```

tobru:
It would be great to see the option "Crossplane Composition (Functions)" discussed in the decision document. This technology could probably also solve the problem at hand. I don't want to say it's the best solution, but at least I'd like to see a discussion around it and a conscious decision why not to use it. Some references to learn more: https://docs.crossplane.io/v1.18/concepts/compositions/ "Compositions are a template for creating multiple managed resources as a single object."

DebakelOrakel:
> It would be great to see the option "Crossplane Composition (Functions)" discussed in the decision document. This technology could probably also solve the problem at hand. I don't want to say it's the best solution, but at least I'd like to see a discussion around it and a conscious decision why not to use it. Some references to learn more: https://docs.crossplane.io/v1.18/concepts/compositions/ "Compositions are a template for creating multiple managed resources as a single object."

tobru:
>  Also Option 1 discusses using existing tools.

Option 1 discusses policy tools, Crossplane isn't one of them.

> I don't see how Crossplane Compositions would be the proper tool for this task.

This is fine, but I want to understand the "why". Crossplane Compositions is exactly that, a template engine which reconciles resources.

This document which is being written serves as a decision document to look back in the future and understand why something was done in a certain way, what alternatives were out there and why they haven't been chosen.

bastjan:
I'll add an option for the 737-MAX 👍

DebakelOrakel:
> Option 1 discusses policy tools, Crossplane isn't one of them.
That is true, the conclusion still remains.

* Crossplane is primarly designed to manage external infrastructure resources.
* Using a full-blown control-plane solution just to do something that a very simple operator can do much more efficiently.
* Instead of using technology we already use (Jsonnet) we would need to learn a very complex framework for a simple task.

Above points are the reason why we never considered using Crossplane, bc it is never was a reasonable choice to consider given the problem statement.

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,502 @@
...
+
+Not every cluster configuration is fully static and can be managed through GitOps.
+
+Some configurations depend on the cluster state or might want to be managed by upstream, by end-users, or by third-party products directly.
+
+We got an extensive array of tools to manage these configurations, from bash "reconcilers," to cron jobs, to policy engines to fully fledged custom controllers.
+Those tools are all specialized and don't cover all use-cases.
+
+A maintainable, fully instrumented, and extensible controller that can manage arbitrary configurations in a cluster is missing.
+We decided to try our hand at creating one.
```
simu, line 13: Maybe link to the decision here:

```suggestion
xref:oc4:ROOT:explanations/decisions/prometheusrule-controller.adoc[We decided] to try our hand at creating one.
```

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,502 @@
...
+
+== High Level Goals
+
+* We can manage arbitrary configurations in a cluster
+** The configuration may depend on external state
```
simu, line 18: maybe clarify external state here? is the state external, but in-cluster? or external behind an arbitrary URL?

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,502 @@
...
+        stringData: {
+          'my-key': 'my-value',
+        }
+      }
+    ]
+----
+<1> Cluster scoped `ClusterManagedResource` can access all resources in the cluster.
+This includes cluster scoped resources.
+<2> The full manifest of the triggering is available.
+<3> A cluster managed resource can create resources in any namespace and thus needs to set the namespace.
```
simu, line 412: this is a bit inconsistent/incomplete, and I'm not sure what the actual example wants to do. I'd like to have a consistent example here (either copy config map to target namespace -- in that case the source is missing, or generating netpols, in which case the full object should be visible in Jsonnet?

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,502 @@
...
+We want to show a notification in the OpenShift Console when a minor upgrade is available.
+To implement this we added a job triggered by an ArgoCD sync.
+The job is reading another resource and creating a console notification manifest from it.
+
+* https://github.com/appuio/component-openshift4-console/blob/740628ebf3822ea82a64fade1e42eb9ff52f67c7/tests/golden/upgrade-notification/openshift4-console/openshift4-console/31_upgrade_notification.yaml#L63[component-openshift4-console ArgoCD sync hook]
+* https://github.com/appuio/component-openshift4-console/blob/740628ebf3822ea82a64fade1e42eb9ff52f67c7/component/scripts/create-console-notification.sh[component-openshift4-console bash script]
+
+=== Sync OpenShift Console TLS Secret with a bash "reconciler"
+
+Cert-manager only allows to create certificates in the same namespace as the ingress.
```
simu, line 427:
```suggestion
Cert-manager only allows to create the secret holding the certificate data  in the same namespace as the `Certificate` resource.
```

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,502 @@
...
+OpenShift Console ingress is in the `openshift-console` namespace, but the secret needs to be applied in the `openshift-config` namespace.
```
simu, line 428: This isn't directly coupled with the OpenShift console ~~ingress~~ route

```suggestion
The OpenShift Console route is in the `openshift-console` namespace, but the secret needs to be applied in the `openshift-config` namespace.
However, we must create the `Certificate` resource in `openshift-console`, since otherwise the OpenShift ingress doesn't admit the HTTP challenge ingress.
```

simu (commented):
Decision LGTM. Architecture LGTM overall, some comments inline.

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,509 @@
...
+        metadata: {
+          name: 'injected-cm',
+          namespace: ns.metadata.name, <3>
+        }
+        data: base.data,
+      }
+    ]
+----
+<1> Cluster scoped `ClusterManagedResource` can access all resources in the cluster.
+This includes cluster scoped resources.
```
simu, line 417:
```suggestion
<1> Cluster scoped `ClusterManagedResource` can access all resources in the cluster.
This includes cluster scoped resources such as `Namespace`.
```

haasad (approved):
Kyverno generate policies have a `synchronize` property that allows you to choose not to sync a resource after it was created. Could something similar be done with espejote?

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
```diff
@@ -0,0 +1,510 @@
...
+
+Currently, we support two types of triggers:
+
+==== `interval`
+
+The template is re-rendered every `interval`.
+
+==== `watchResource`
+
+The template is re-rendered whenever the watched resource is created, changed, or deleted.
```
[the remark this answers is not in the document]
bastjan: That's a good idea! `watchResource` is the wrong context because it's input only and we do not really know what the underlying event was.

I'll add `applyOptions.createOnly` to use for only creating the resulting resource from the template.

docs/modules/ROOT/pages/references/architecture/espejote-in-cluster-templating-controller.adoc
bastjan: clarified as `state managed by other controllers inside the cluster`

PR #370, opened 2024-12-06
Decision for autoscaling with restricted downscaling

simu:


haasad (approved):
Decision looks fine!

Two thoughts that came up while reviewing this:
- How big is the chance/risk of two nodes being drained simultaneously? One by the maintenance and one for downscaling. This could create problems for deployments without PDBs that rely on the fact the we usually never drain two nodes at the same time.
- Is there a risk that the autoscaler will always add a node during the maintenance (because resources are tight during node rebooting) which will then not be downscaled anymore and lay around until the next maintenance?

PR #379, opened 2025-02-04
Decision: Managed Load Balancer and Autoscaling cloudscale Infra Nodes

bastjan:


simu (approved):
LGTM and we can workaround the issue with internal traffic on K8s < 1.30 by adding the following annotation to the LB service:

```
k8s.cloudscale.ch/loadbalancer-force-hostname: ingress.$CLUSTER_DOMAIN
```

PR #388, opened 2025-02-18
Add how-tos for managing nodes for Exoscale clusters with instance pools

simu:
* How-to for removing a node on an Exoscale cluster with instance pools
* How-to for changing node type on an Exoscale cluster with instance pools
* Some formatting cleanup in the partials used by the Exoscale node how-tos

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/exoscale/remove_node_instancepool.adoc
+++ b/docs/modules/ROOT/pages/how-tos/exoscale/remove_node_instancepool.adoc
@@ -1,0 +1,98 @@
+= Remove a worker node (instance pool)
+
+:cloud_provider: exoscale
+:kubectl_extra_args: --as=cluster-admin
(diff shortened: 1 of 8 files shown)


docs/modules/ROOT/pages/how-tos/exoscale/remove_node_instancepool.adoc
```diff
@@ -0,0 +1,98 @@
...
+== Starting situation
+
+* You already have a OpenShift 4 cluster on Exoscale
+* Your cluster uses instance pools for the worker and infra nodes
+* You have admin-level access to the cluster
+* You want to remove an existing worker node in the cluster
+
+== High-level overview
+
+* Drain the node
```
bastjan, line 25:
```suggestion
* We drain the node
```

To make it analogous to the other steps.

bastjan (approved):

PR #402, opened 2025-04-15
Update cloudscale install instructions to provision cloudscale LB for ingress

simu:
* Extract ingress floating IP from Terraform state and configure in cluster config
* Deploy our cloudscale-loadbalancer-controller (https://github.com/appuio/cloudscale-loadbalancer-controller/)
* Deploy cert-manager before the cloudscale-loadbalancer-controller
* Various cleanups for steps that are no longer necessary


docs/modules/ROOT/partials/install/bootstrap-nodes.adoc
```diff
@@ -168,6 +202,25 @@ kubectl apply -f ../machine-api-provider-cloudscale/11_deployment.yaml
 ----
 kubectl apply `for f in ../openshift4-nodes/machineset-*.yaml ; do echo -n "-f $f " ; done`
 ----
+
+. Deploy the cloudscale loadbalancer controller
++
+[source,bash]
+----
+kubectl apply -f ../cloudscale-loadbalancer-controller/00_namespace.yaml
+kapitan refs --reveal --refs-path ../../refs -f ../cloudscale-loadbalancer-controller/10_secrets.yaml | kubectl apply -f -
```
bastjan, line 211: I wonder if we need a disclaimer that commodore needs to be installed with `uv` to get kapitan in the path?

PR #404, opened 2025-04-25
Update install instructions for OpenShift 4.18 on Exoscale

thobens:


simu:
outdated, we've now updated the instructions to 4.19 in #427

PR #417, opened 2025-07-01
Update cloudscale install instructions

simu:
* Remove some unnecessary configs
* Further simplify infrastructure provisioning (reordering some steps, removing some steps that aren't required anymore)
* Create image registry and logstore buckets with `mc`
* Add command to create a copy of the default `UpgradeConfig` of the installed cluster to trigger a first maintenance

TODO: Figure out if we need to add the image registry changes for other cloud providers as well


docs/modules/ROOT/partials/install/prepare-syn-config-terraform.adoc
```diff
@@ -49,15 +49,6 @@ yq eval -i ".parameters.openshift4_terraform.terraform_variables.rhcos_template
...
 ----
 
-endif::[]
-ifeval::["{provider}" == "cloudscale"]
-. Configure cloudscale.ch-specific Terraform variables
-+
-[source,bash,subs="attributes+"]
-----
-yq eval -i ".parameters.openshift4_terraform.terraform_variables.image_slug = \"custom:rhcos-{ocp-minor-version}\"" \
-  ${CLUSTER_ID}.yml
```
simu, line 59: We now set this variable in https://git.vshn.net/syn/commodore-defaults/-/blob/5de93dc03471c160e3887197db140745584ab93b/distribution/openshift4/cloud/cloudscale/machinesets.yml#L56 (internal)

PR #419, opened 2025-07-08
Architecture document for SLI reporting service

HappyTetrahedron:


docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
```diff
@@ -0,0 +1,216 @@
...
+A simple query API can be used to query for the updated SLOs.
+
+The service runs on VSHN's centralized infrastructure, same as VSHN's central monitoring.
+Thanks to this, the service doesn't need to be exposed to the internet.
+
+
+=== Service architecture
+
+The SLO reporting service consists of two architectural components: the downtime window API, and the data querying API.
+While both components are largely independent and could easily be split into separate services, we gain no real benefit from doing so: we require none of the benefits provided by a microservice-style architecture.
```
HappyTetrahedron, line 40: I am actually a bit on the fence whether this makes sense to keep as a monolithic service; the two separate functions are really very independent.

My current reasoning for keeping them together is easier development & deployment. But an argument could be made that if they're separate services, they could be developed in parallel and each part could be swapped out more easily (particularly the querying part might change in the future).

I know how yall like "neat" architecture. Wdyt? ^^

docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
```diff
@@ -0,0 +1,216 @@
...
+            }
+        )})
+    }
+)}).*, function($l,$r){
+    $l.timestamp > $r.timestamp
+})
+----
+
+Since it's only possible to query one cluster at a time, it's not possible to create single panels that show data for all clusters on the same graph.
+However, it's possible to create a repeat panel that automatically generates one panel per cluster, which should be sufficient for an overview.
```
HappyTetrahedron, line 213: I figured this is good enough. Wdyt?

docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+The SLO reporting service provides a simple querying API to retrieve adjusted SLO data for a specific cluster and time window.
+The querying API is effectively a proxy for the Mimir querying API.
+
+The querying API is capable of providing the following for each customer-facing SLO:
+
+* Adjusted SLI Error Rate: this corresponds to `slo:sli_error:ratio_rate1h` with every data point inside a relevant downtime window set to zero.
+* Adjusted Cumulative SLI Error Rate: This corresponds to a `sum_over_time( .. [32d:1h])` of the above result.
+* Adjusted Cumulative SLO Burndown: This corresponds to 1 minus (the above divided by the corresponding SLO error budget).
+* Adjusted Cluster Availability Percentage: This corresponds to the SLO plus (1-SLO times the above).
```
HappyTetrahedron, line 96: These are all the metrics we use in current dashboards. Metrics 2, 3 and 4 can be derived from the first, though, if we also know the SLO objective and the error budget.
You'll find that further down in my sample query response, instead of reporting all 4 of these metrics, I just report metric 1 (the error rate) and the objective and budget. The other metrics are then calculated on the fly in Grafana.

I'm not sure what's best here. There are two possibilities:
1) Calculate all derived metrics in the service, just use Grafana to display them
2) Only calculate the basic metric in the service, use Grafana to calculate any derivative metrics.

With 1), we have the calculations centralized, but chances are that we need to update the service whenever we want a new dashboard that displays the data slightly differently. And chances are that we use Grafana to further mangle and transform the data anyway - because Grafana needs it in very specific formats depending on the visualization - so we'll likely end up with both transformations in the Service and transformations in Grafana. At that point we might as well put everything into Grafana. If we want to go the other way and put everything into the Service, then the service must by necessity become very Grafana-aware and provide its data in very specific (insane) formats.
With 2), we gain flexibility to adjust the derivative metrics on the fly in Grafana, but we'll end up with a lot of complex logic hidden in Grafana panel configurations.

An example of how option 2 would work on the Grafana side is provided later (essentially, we use jsonata to do all the transformations; jsonata is pretty turing-complete so we will likely never run into real limitations for what is possible to do, but it gets pretty cursed.)

docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
```diff
@@ -0,0 +1,216 @@
...
+Downtime windows can be added manually via the Downtime Window API.
+However, in some cases we can automate the reporting of downtime windows, such as when the 3rd party maintains a status page.
+Generally, such status pages provide an RSS feed with information about downtimes.
+
+RSS feeds are only loosely standardized and it's hard to generalize a field mapping for this specific use of RSS.
+For that reason, it would be significant effort to build our own RSS reading service.
+Instead, RSS ingestion can be handled by a simple bash script that's regularly run in a Kubernetes CronJob on VSHN's central cluster.
+
+We can use multiple bash scripts for different sources of RSS feeds.
+The bash script sends the parsed outage time window to the Downtime Window API.
```
HappyTetrahedron, line 173: Is this good enough? I wasn't sure how many statuspages we'd be monitoring this way, I'm assuming it's only a small handful.

RSS is pretty cursed to work with; not looking forward to this part.

It is possible that some general-purpose RSS transformer already exists, which we can use. (I have yet to check)

docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
simu: we're already doing a repeat panel for the current "Cluster availability" dashboard, I don't see a need to do something else.

docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
simu: I'm not sure what's actually better here, but I would make sure that we provide the base metrics (1h error rate, objective and error budget) regardless of which other metrics we pre-compute in the service. That way we can still experiment a bit with adjusted queries in Grafana and have the option to lift new computations into the service once we've convinced ourselves that the adjusted computation is "better".

As you mention, I expect that we will need to do some transformations in Grafana anyway, but I think it's valuable to have some of the more complex computations in a code base where we can write tests for them.

HappyTetrahedron:
I've added the missing pieces of the design doc, so it's now up for "proper" review

docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
```diff
@@ -0,0 +1,322 @@
...
+Any parameters marked as optional may be omitted.
+
+If an `ExternalID` is provided in the request body, and an existing downtime window shares the same `ExternalID`, then the existing window should be updated (replaced) instead of a new one created.
+Otherwise, a new record is created from the request body and assigned a new random ID.
+
+Returns:
+
+The newly created record, including ID.
+
+====== `POST /downtime/:id`
```
simu, line 154: If we want to keep `POST` as a proper POST, we should also support `PATCH` for partial updates of existing downtime records (e.g. just adjusting start or end time).

simu (commented):
Architecture LGTM overall. See inline comments regarding API endpoints.

docs/modules/ROOT/pages/references/architecture/sli_reporting.adoc
```diff
@@ -0,0 +1,341 @@
...
+
+Updates an existing downtime window, supporting partial updates.
+
+Path parameters:
+* `id`: ID of the existing downtime window record to update.
+
+Body parameters:
+
+The body of the request corresponds to the JSON example given above.
+Any parameters marked as optional may be omitted.
```
simu, line 183: For `PATCH` it should be possible to omit all parameters (so we can just patch the start **or** end of a downtime for example), otherwise it's exactly the same as `POST`.

PR #473, opened 2026-03-26
Decision: Backup OpenShift Image Registry

DebakelOrakel:


docs/modules/ROOT/pages/explanations/decisions/image-registry-backup.adoc
```diff
@@ -0,0 +1,63 @@
...
+We provide backups for cluster objects, but if a customer uses the OpenShift Image Registry as only registrt for their images we need a way to backup that as well.
+
+=== Goals
+
+* Have the means to restore the OpenShift Image Registry if required.
+
+
+== Proposals
+
+=== Option 1: Use PVC backed Image Registry and K8up
```
megian: From my point of view: This is not really an option. You can not guarantee the consistency of an image and the end result is a Schrödingers backup. You will never know if a restored container registry is recoverable and finally constant. If you can't guarantee the recovery the hole process is useless.

megian (changes requested):

simu (approved):

PR #479, opened 2026-04-23
Cluster Access Tooling Architecture Docs

bastjan:
This PR proposes tooling to easen cluster access over VSHN jumphosts.

The proposal includes a SOCKS5 proxy routing connections, and cli tooling.
The proxy is designed to stay in the background and be transparent. CLI tools, such as `curl`, and browsers can statically add the proxy as a hop.

The tooling gets all information from VSHN SSHOP config and the Lieutenant inventory.

Most technical data was verified in the PoC: https://github.com/vshn/kharon

### DIFF
--- a/docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
+++ b/docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
@@ -1,0 +1,433 @@
+= Cluster Access Tooling
+
+[abstract]
+--
+This document proposes a new cluster access tooling which aims to provide a user-friendly solution to access our managed OpenShift clusters without a public API.
+
+The proposal includes a SOCKS5 proxy routing connections, and CLI tooling.
+The proxy is designed to stay in the background and be transparent. CLI tools, such as `curl`, and browsers can statically add the proxy as a hop.
+
+The tooling gets all required information from VSHN SSHOP config and the Lieutenant inventory.
+--
+
+== Problem Statement
+
+A growing part of our managed OpenShift clusters are in private networks without direct access from the public internet.
+Those clusters are accessed through `ssh` jump hosts that proxy connections to the cluster using dynamic port forwarding (`ssh -D`).
+
+The current solution is to, sometimes automatically using direnv, set up a SOCKS proxy per cluster and then use that proxy to access the cluster using `oc` or `kubectl`.
+Since each proxy runs on a different local port, users have to keep track of which proxy corresponds to which cluster.
+In Firefox this routing is done using the awesome `FoxyProxy` addon, in other applications users have to set up routing rules manually.
+
+With the growing number of extra services sold on top of our managed OpenShift clusters, the number of other teams and users within VSHN needing access to the clusters is growing.
+The solution works well enough for our small team, but it's not very user-friendly if you don't use it on a daily basis.
+
+We need a simple, user-friendly, and scalable solution to access our managed OpenShift clusters in private networks.
+
+== High Level Goals
+
+* An easy-to-use tool to access clusters in private networks.
+* Users should be able to access clusters in private networks without fiddling with Firefox addons or manual routing rules.
+* The solution should integrate well with our existing `ssh` jump host setup.
+
+== Non-Goals
+
+* Replacing `ssh` jump hosts with a different solution.
+* Replacing cluster authentication.
+
+== Constraints
+
+* The solution must work with our existing `ssh` jump host setup.
+* No additional tunnels or external connections through customer firewalls.
+  Additional tunnels might not be approved by customers and would further increase the sprawl of access solutions VSHN has to maintain.
+
+== Interfaces
+
+=== User
+
+The user interacts with the cluster access tooling through a command line interface.
+
+=== Routing information
+
+Routing information is queried from our central inventory, which is managed using https://syn.tools/syn/index.html[Project Syn] and can be queried through the https://syn.tools/lieutenant-api/references/index.html[Lieutenant API].
+
+See <<routing-information-retrieval>> for more information.
+
+=== SSH configuration
+
+The tooling uses the user's SSH configuration.
+https://vshnwiki.atlassian.net/wiki/spaces/VT/pages/8291275/SSH+Jumphost+sshop[SSH Jumphost (sshop)] configuration is included in the user SSH configuration.
+
+The tooling reads the user's SSH configuration to determine how to connect to the jump hosts and clusters.
+
+=== HTTP based access
+
+The tooling sets up a local SOCKS5 proxy listening on localhost.
+
+The SOCKS5 proxy is transparent to the user, non-routed traffic is directly connected to the destination, while traffic to cluster endpoints is routed through the proxy.
+Direct connections shouldn't affect speed or battery life, as they should be handed off to the kernel without going through the proxy.
+
+The proxy can stay running in the background.
+
+=== VPNs and other network access solutions
+
+We decided to not try to deal with VPNs.
+There is no stable cross-platform solution for managing the VSHN VPN.
+
+Clusters that require VPN access are routed through an additional VSHN management jump host that has access to the VPN, so users don't have to worry about setting up the VPN themselves.
+
+==== OpenShift CLI (`oc`) / Kubernetes CLI (`kubectl`) / curl / other CLI tools
+
+The user must set shell environment variables `*_PROXY` to use to access the clusters in private networks.
+
+[source,bash]
+----
+export HTTP_PROXY=socks5h://localhost:12000 <1>
+export HTTPS_PROXY=socks5h://localhost:12000
+
+curl https://api.c-bettersmarter-prod01.vshnmanaged.net:6443
+----
+<1> `socks5h` is used to ensure that DNS resolution is also done through the proxy, which is necessary for non-public DNS names.
+The port is an example, the actual port is configurable.
+
+==== Firefox and other web browsers
+
+The user must configure their web browser to use the SOCKS5 proxy to access the clusters in private networks.
+
+Because the proxy can stay running in the background, the browser can be configured to use it permanently, so users don't have to fiddle with browser settings every time they want to access a cluster in a private network.
+
+DNS should also be routed through the proxy to ensure that cluster endpoints that aren't publicly resolvable can be accessed.
+Firefox can be configured to do this by checking the "Proxy DNS when using SOCKS v5" option in the proxy settings.
+
+=== Cluster Authentication
+
+Cluster authentication isn't handled by the cluster access tooling, but the tooling should work well with our existing cluster authentication solutions.
+
+==== Emergency Credentials
+
+xref:oc4:ROOT:references/architecture/emergency_credentials.adoc[] should be integrated into the cluster access tooling to allow users to easily retrieve emergency credentials for clusters in private networks.
+
+== Implementation
+
+The cluster access tooling is implemented as a command line tool written in Go.
+Go allows cross-compilation, static binaries, and has good support for SSH and HTTP.
+
+It can run as a daemon in the background to keep the SOCKS5 proxy running, or interactively as a CLI.
+The interactive CLI can be used, among other things, to update routing information, test jump host connectivity, or retrieve emergency credentials.
+
+=== SOCKS5 Proxy [[socks5-proxy]]
+
+The tool sets up a local SOCKS5 proxy on localhost that routes traffic to cluster endpoints through the appropriate jump hosts based on the routing information retrieved from the Lieutenant API.
+All other traffic is directly connected to the destination.
+The implementation should hand off non-routed traffic to the kernel to avoid unnecessary overhead.
+
+The list of routed endpoints should be searched by suffix so that subdomains of cluster endpoints are also routed through the proxy without needing to list every possible subdomain in the routing information.
+The list should be searched by longest suffix first to allow for more specific routing rules to take precedence over more general ones.
+
+Exclusion rules take precedence over inclusion rules, so if a domain matches both an inclusion and an exclusion rule, it should be routed directly to the destination instead of through the proxy.
+
+[source]
+----
+c-bettersmarter-prod01.vshnmanaged.net
+a.storage.bettersmarter.ch
+b.storage.bettersmarter.ch
+vcenter.bettersmarter.ch
+----
+
+A POC implementation of the SOCKS5 was able to demonstrate non-noticeable overhead for new TCP connections, and zero overhead for established connections.
+
+An optional username and password can be set for the SOCKS5 proxy to prevent unauthorized access to the proxy.
+We don't see it as necessary to add authentication by default, as the proxy is only listening on localhost, and only does tunneling to cluster endpoints, so it shouldn't be a security risk.
+
+=== SSH Tunneling
+
+SSH connections are set up when first required and kept alive as long as the proxy is running to avoid the overhead of setting up new SSH connections for every request.
+Connections are monitored and automatically re-established if they drop due to network issues or if the machine goes to sleep.
+
+The tooling uses the `golang.org/x/crypto/ssh` package to set up SSH tunnels to the jump hosts and clusters.
+This allows us to have full control over the tunnels, we don't have to worry about managing `ssh` processes and ports, and we can easily keep the tunnels alive in the background.
+VSHN SSH configuration is pretty standardized, so we can implement support for the features we use without needing to support every possible SSH configuration option.
+If a user has a complex SSH configuration that the tooling doesn't support, we can throw an error early to avoid unexpected behavior.
+
+We evaluated using the `ssh` command line tool with `-D` for dynamic port forwarding.
+The main advantage of using the `ssh` command line tool is that it can use the user's existing SSH configuration without having to re-implement all the SSH features in Go.
+The main advantage of supporting all features is also the biggest disadvantage, as it can lead to unexpected behavior if the user has a complex SSH configuration that the tooling doesn't support.
+Another issue is the port management, as each `ssh -D` command requires a free local port, which can lead to conflicts and makes it harder to manage the proxies.
+Keeping the tunnel alive through the command line tool can be difficult, especially with `ControlMaster` and `ControlPersist` options.
+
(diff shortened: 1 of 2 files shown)


DebakelOrakel (approved):

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,262 @@
...
+==== `test`
+
+The `test` sub-command tests the connectivity to all jump hosts and clusters based on the routing information retrieved from the Lieutenant API.
+This can be used to verify that the routing information is correct and that the jump hosts and clusters are accessible.
+
+==== `clusters`
+
+The `clusters` sub-command lists the clusters based on the information retrieved from the Lieutenant API.
+
+==== `clusters CLUSTER_NAME kubeconfig`
```
simu, line 347: Might be interesting to also have a variant of this command which generates kubeconfigs for all known clusters (based on Lieutenant API?) or a single kubeconfig with a context for each cluster. I'm not picky about the exact format (personally I want separate kubeconfigs per cluster, but I already have https://github.com/simu/kubectl-split-config to generate per-cluster kubeconfigs from a big kubeconfig with many contexts)

Also, this probably needs more information than the facts currently outlined in this document (especially for clusters that don't need a jumphost).

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,262 @@
...
+
+The `clusters CLUSTER_NAME kubeconfig` sub-command retrieves the kubeconfig for the specified cluster and writes it to the specified file or standard output if no file is specified.
+
+==== `clusters CLUSTER_NAME emergency-credentials`
+
+The `clusters CLUSTER_NAME emergency-credentials` sub-command retrieves the emergency credentials kubeconfig for the specified cluster and writes it to the specified file or standard output if no file is specified.
```
simu, line 353: Same here: for seamless use, this command needs to know the API URL of each cluster -> we probably should add that as a fact (either as a fact during cluster setup, or through Steward's additional dynamic facts configmap, since we already have the API URL for OCP clusters in the inventory).

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,389 @@
...
+[source,yaml]
+----
+apiVersion: syn.tools/v1alpha1
+kind: Cluster
+metadata:
+  name: c-bettersmarter-prod01
+spec:
+  displayName: BetterSmarter Prod01
+  facts:
+    apiURL: https://api.c-bettersmarter-prod01.vshnmanaged.net:6443 <1>
```
bastjan: Not sure if we want to make it a dynamic fact? It feels right in the cluster facts and feels bit bloated to add it with a rountrip through compile->argocd->steward. *But* we already have it in the catalog.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,389 @@
...
+    jumphost: management4.rma1.bettersmarter.vshnmanaged.net <2>
+    jumphostDomains: c-bettersmarter-prod01.vshnmanaged.net,vcenter.bettersmarter.ch,storage.bettersmarter.ch <3>
+    noJumphostDomains: apps.c-bettersmarter-prod01.vshnmanaged.net <4>
```
bastjan: Better naming welcome 🥺

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
simu: No hard feelings either way, I suggested the dynamic fact because that would reuse the current source of truth parameters for the cluster API URL etc.

If we go with static facts, We should also consider the following approach: add facts that mimic the current `openshift` parameters (and replace the default values for those with the appropriate facts):

Currently (global defaults):

```
  openshift:
    baseDomain: ${cluster:name}.example.com
    appsDomain: apps.${openshift:baseDomain}
    apiURL: api.${openshift:baseDomain}
```

After (Lieutenant):

```
facts:
  openshiftBaseDomain: c-bettersmarter-prod01.vshnmanaged.net
  openshiftAppsDomain: apps.${facts:openshiftBaseDomain}
  openshiftApiURL: api.${facts:openshiftBaseDomain}
```

and (global defaults)

```
openshift:
  baseDomain: ${facts:openshiftBaseDomain}
  appsDomain: ${facts:openshiftAppsDomain}
  apiURL: ${facts:openshiftApiURL}
```

I've quickly verified that using references in facts works fine in Commodore (would need special handling in other tools though). Overall, it's not super clean either, because we now need to set 3 additional facts for each new cluster, two of which usually will be left as default.

If we go with static facts, I'd omit the API scheme and port in the fact since it's always 6443 for OpenShift and if either are included in the fact, we can't use the fact to set parameter `openshift.apiURL` which is used in places where we don't want the port and scheme.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
simu: not sure if it's better, but we could mimic the Linux `no_proxy` naming:

```suggestion
    proxyDomains: c-bettersmarter-prod01.vshnmanaged.net,vcenter.bettersmarter.ch,storage.bettersmarter.ch <3>
    noProxyDomains: apps.c-bettersmarter-prod01.vshnmanaged.net <4>
```

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,407 @@
...
+    openshiftOAuthRoute: '"oauth-openshift.apps.c-bettersmarter-prod01.vshnmanaged.net"' <5>
+----
+<1> The API URL of the cluster, which is used for the generated kubeconfig.
+<2> The hostname of the jump host to use for accessing the cluster.
+Only the last host in the jump host chain should be specified, the tooling looks up the routing information for the jump host by querying the SSH configuration.
+<3> Comma separated list of additional domains that should be routed through the same jump host.
+This can be used for access to management interfaces such as the vCenter interface the cluster is running on.
+<4> Comma separated list of domains that shouldn't be routed through the jump host.
+This can be used for access to cluster endpoints that aren't accessible from the DMZ.
+<5> Already existing dynamic fact that can be used for `oc login`.
```
simu: Afaik, `oc login` doesn't need the oauth route, you just point it at the cluster's API URL.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,407 @@
...
+[%collapsible]
+====
+[source,ini]
+....
+# ~/.config/systemd/user/smart-access.service
+[Unit]
+Description=Smart Access cluster proxy
+
+[Service]
+Environment="SSH_AUTH_SOCK=%t/gcr/ssh" <1>
```
simu, line 255: ~~Most likely this will not work for everyone out of the box.~~

After some more testing, this might work on all systemd-based distros (including auto key unlock with PAM).

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,407 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func main() {
+	kubeconfig := "/tmp/kubeconfig"
```
simu: for a real version this should probably be a cluster-specific path, otherwise running multiple `clusters CLUSTER_NAME shell` for different clusters in parallel will create chaos.

bastjan: True, my intentions was always something like `$XDG_CACHE_HOME/tool/c-cluster1/kubeconfig`. Will add a callout.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
bastjan: I only tested on Gnome based systems to be honest.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
simu: New idea from discussion with @HappyTetrahedron and @bastjan: go with dynamic facts for OpenShift base domain, apps domain, API URL and console URL. This doesn't require a new source of truth for the values (we keep the source of truth in inventory parameter `openshift`), doesn't require custom rendering of reclass references outside Commodore, and allows us to generate a default `jumphostDomains` set from the dynamic facts.

Additionally, having these dynamic facts will be generally useful for other integrations that consume data from the Lieutenant API (e.g. Gandalf workflows) or Lieutenant metrics (e.g. Grafana dashboards).

Roughly, we'd have something like the following:

```
status:
  facts:
    openshiftBaseDomain: c-bettersmarter-prod01.vshnmanaged.net
    openshiftAppsDomain: apps.c-bettersmarter-prod01.vshnmanaged.net
    openshiftApiURL: api.c-bettersmarter-prod01.vshnmanaged.net
    openshiftConsoleURL: console.apps.c-bettersmarter-prod01.vshnmanaged.net
```

The (naively) generated `jumphostDomains` would be `['c-bettersmarter-prod01.vshnmanaged.net', 'apps.c-bettersmarter-prod01.vshnmanaged.net', 'api.c-bettersmarter-prod01.vshnmanaged.net', 'console.apps.c-bettersmarter-prod01.vshnmanaged.net']`. While this has a few redundant entries, this accounts for any custom values for apps domain, API URL and console URL.

With this approach, the values in (probably now optional) static fact `jumphostDomains`/`jumphostAdditionalDomains` will be added to the generated list.

Not sure right now whether we'd include scheme and port for API and console URLs with this approach, since we probably don't want those for the actual proxy routing.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
simu: based on discussion with @HappyTetrahedron and @bastjan 

```suggestion
    jumphostDomains: c-bettersmarter-prod01.vshnmanaged.net,vcenter.bettersmarter.ch,storage.bettersmarter.ch <3>
    jumphostSkipDomains: apps.c-bettersmarter-prod01.vshnmanaged.net <4>
```

or with the discussion above about generating a default `jumphostDomains`:

```suggestion
    jumphostAdditionalDomains: vcenter.bettersmarter.ch,storage.bettersmarter.ch <3>
    jumphostSkipDomains: apps.c-bettersmarter-prod01.vshnmanaged.net <4>
```

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
bastjan: > Not sure right now whether we'd include scheme and port for API and console URLs with this approach, since we probably don't want those for the actual proxy routing.

If we name it URL we should add them. This would also allow integrations if the ports ever change. It's rather easy in most languages to parse URLs.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
simu: Yes, after some more research, GCR is part of Gnome Cryptography services, so non-Gnome environments may work differently.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,425 @@
+= Cluster Access Tooling
+
+[abstract]
+--
+A proposal for a new cluster access tooling to provide a user-friendly solution to access our managed OpenShift clusters without a public API.
```
simu:
```suggestion
This document proposes a new cluster access tooling which aims to provide a user-friendly solution to access our managed OpenShift clusters without a public API.
```

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,425 @@
...
+
+The proposal includes a SOCKS5 proxy routing connections, and cli tooling.
```
simu:
```suggestion
The proposal includes a SOCKS5 proxy routing connections, and CLI tooling.
```

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,425 @@
...
+See <<socks5-proxy>> for more information.
+
+==== `test`
+
+The `test` sub-command tests the connectivity to all jump hosts and clusters based on the routing information retrieved from the Lieutenant API.
+This can be used to verify that the routing information is correct and that the jump hosts and clusters are accessible.
+
+==== `kubeconfig`
+
+The `kubeconfig` sub-command generate a kubeconfig for all known clusters based on the routing information retrieved from the Lieutenant API and writes it to the specified file or standard output if no file is specified.
```
simu, line 335: Follow-up question here: will this cover all clusters or just all clusters that match `.facts.distribution=="openshift4" or .facts.distribution=="oke"`?

Or just clusters that actually have the `openshiftApiURL` dynamic fact?

simu (approved):
LGTM. Some final comments inline.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
bastjan: Added a note https://github.com/appuio/openshift4-docs/pull/479/changes#diff-49442fa71adf71163820140781121d06e0e60bde92e763f9507bf06264e62f98R336-R340

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,431 @@
...
+include:
+- console.apps.c-bettersmarter-prod01.vshnmanaged.net
+- apps.c-bettersmarter-prod01.vshnmanaged.net
+- api.c-bettersmarter-prod01.vshnmanaged.net
+- c-bettersmarter-prod01.vshnmanaged.net
+- storage.bettersmarter.ch
+- vcenter.bettersmarter.ch
+----
+
+Information downloaded from the Lieutenant API is cached locally to allow the tooling to continue working even if the Lieutenant API is temporarily unavailable.
```
haasad, line 212: Will this be in a accessible location and format for editing temporarily? I sometimes find myself in a situation where I quickly want to edit `jumphostDomains`, i.e. edit the foxyproxy settings in our current setup.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,431 @@
...
+See <<socks5-proxy>> for more information.
+
+==== `test`
+
+The `test` sub-command tests the connectivity to all jump hosts and clusters based on the routing information retrieved from the Lieutenant API.
+This can be used to verify that the routing information is correct and that the jump hosts and clusters are accessible.
+
+==== `kubeconfig`
+
+The `kubeconfig` sub-command generate a kubeconfig for all known clusters based on the routing information retrieved from the Lieutenant API and writes it to the specified file or standard output if no file is specified.
```
haasad, line 335: I lack some basic understanding here; does creating a kubeconfig automatically imply a login to the cluster? Can I run kubectl commands immediately after running this or does it still require authentication first?

haasad (commented):
Very nice 👏

haasad (approved):

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
simu: My understanding for `kubeconfig` and `clusters CLUSTER_NAME kubeconfig` is that the commands simply generate a kubeconfig file which you can then use by pointing `kubectl`/`oc` to it by setting the `KUBECONFIG` env variable.

If you want to directly do stuff, you're probably looking for the `clusters CLUSTER_NAME shell` command.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
```diff
@@ -0,0 +1,431 @@
...
+See xref:oc4:ROOT:references/architecture/emergency_credentials.adoc[] for implementation details of the emergency credentials retrieval.
+
+==== `clusters CLUSTER_NAME shell`
+
+[NOTE]
+----
+We might try auto-login to the cluster using `oc login` before opening the shell.
+----
+
+The `clusters CLUSTER_NAME shell` sub-command opens a shell with the kubeconfig set up to access the specified cluster.
```
simu, line 364: Idea here: have an option to let the user specify a shell script which gets sourced before the shell is handed to the user -> allows doing stuff like customizing PS1 to the users preference (or run a suitable login command without having to bake that into the tool).

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
bastjan: Yes most probably `XDG_CACHE_DIR` on linux.

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
bastjan: Or at least an environment variable a `.bashrc` or similar could pick up on 👍

docs/modules/ROOT/pages/references/architecture/cluster_access_tooling.adoc
bastjan: > does creating a kubeconfig automatically imply a login

No, that's out of scope as there's many ways of logging into a kubernetes cluster.

I will implement an additional OCP specific command.

